PR #13, opened 2024-02-15
Add release pipeline

movsic:


.github/scripts/get_min_versions.py
```diff
@@ -0,0 +1,65 @@
+import sys
+
+import tomllib
+from packaging.version import parse as parse_version
+import re
+
+MIN_VERSION_LIBS = ["langchain-core", "langchain-community", "langchain"]
```
lkuligin, line 7: do we need this?

.github/workflows/_release.yml
```diff
@@ -0,0 +1,282 @@
...
+        run: make tests
+        working-directory: ${{ inputs.working-directory }}
+
+      - name: 'Authenticate to Google Cloud'
+        id: 'auth'
+        uses: google-github-actions/auth@v2
+        with:
+          credentials_json: '${{ secrets.GOOGLE_CREDENTIALS }}'
+
+      - name: Run integration tests
```
lkuligin, line 166: let's disable this maybe until we don't migrate to new integration testing component?

.github/scripts/get_min_versions.py
efriis: Good to test for core

.github/workflows/_release.yml
efriis: Would prefer to keep integration tests running before releases! What is the new component? New script?

efriis (approved):

PR #16, opened 2024-02-19
Fix no args function call

jzaldi:
- Fixed Key Error 'required' when defining a tool with no arguments.
- Added unit test for function utils

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/functions_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/functions_utils.py
@@ -1,5 +1,5 @@
 import json
-from typing import Dict, List, Type, Union
+from typing import Any, Dict, List, Type, Union
 
 from langchain_core.exceptions import OutputParserException
 from langchain_core.output_parsers import BaseOutputParser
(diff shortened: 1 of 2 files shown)


libs/vertexai/langchain_google_vertexai/functions_utils.py
```diff
@@ -54,7 +54,7 @@ def _format_tool_to_vertex_function(tool: BaseTool) -> FunctionDescription:
                     }
                     for k, v in schema["properties"].items()
                 },
-                "required": schema["required"],
+                "required": schema.get("required", []),
```
lkuligin, line 57: nits: should we maybe just remove `requried` arg if it's empty? it looks like it's not a required one

lkuligin, line 57: should we update line #34 too?

jzaldi: Wraped both blocks into a function, as it was duplicated code and make it so that if there is no required there's no key

lkuligin (approved):

PR #17, opened 2024-02-19
Refactor matching_engine (migration from lc monorepo)

jzaldi:
Modular implementation of `VectorStore` with VertexAI components.
Migration from PR to LangChain MonoRepo

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
+++ b/libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
@@ -1,0 +1,258 @@
+import uuid
+from typing import TYPE_CHECKING, Any, Iterable, List, Optional, Tuple, Type, Union
+
+from google.cloud.aiplatform.matching_engine.matching_engine_index_endpoint import (
+    Namespace,
+)
+from langchain_core.documents import Document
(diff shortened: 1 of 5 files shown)


libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
```diff
@@ -0,0 +1,258 @@
...
+        """This function returns the default embedding.
+        Returns:
+            Default TensorflowHubEmbeddings to use.
+        """
+
+        # TODO: Change to vertexai embbedingss
+
+        from langchain_community.embeddings import TensorflowHubEmbeddings
+
+        return TensorflowHubEmbeddings()
```
lkuligin, line 197: should we maybe add a deprecation warning and switch to different default embeddings (VertexAI) so that we can drop dependency on `langchain_community` completely?

RajeshThallam:
@jzaldi Does the [doc/notebook](https://github.com/langchain-ai/langchain/blob/master/docs/docs/integrations/vectorstores/google_vertex_ai_vector_search.ipynb) need to be updated or changes are backward compatible?

lkuligin:
updating external docs will follow after the release to pypi

lkuligin (approved):

PR #22, opened 2024-02-21
Added support for imagetext models

jzaldi:
Added support fort `imagetext` models

- `VertexAIImageCaptioning` LLM based implementation of image captioning
- `VertexAIImageCaptioningChat` Chat based implementation of image captioning
- `VertexAIVisualQnAChat` Chat based implementation of image question and answering

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vision_models.py
+++ b/libs/vertexai/langchain_google_vertexai/vision_models.py
@@ -14,7 +14,7 @@ from vertexai.vision_models import Image, ImageTextModel  # type: ignore[import-
 from langchain_google_vertexai._image_utils import ImageBytesLoader
 
 
-class BaseImageTextModel(BaseModel):
+class _BaseImageTextModel(BaseModel):
     """Base class for all integrations that use ImageTextModel"""
 
     model_name: str = Field(default="imagetext@001")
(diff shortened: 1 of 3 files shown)


libs/vertexai/langchain_google_vertexai/vision_models.py
```diff
@@ -0,0 +1,276 @@
...
+            string representation of the image.
+                Currently supported are:
+                    - Google Cloud Storage URI
+                    - B64 encoded string
+                    - Local file path
+                    - Remote url
+                There has to be at least other message with the first question.
+        """
+
+        if len(messages) < 2:
```
lkuligin, line 212: I don't think we should expect two messages, I'd rather suggest we should expect one with multiple content parts:
```
image_message = {
        "type": "image_url",
        "image_url": {"url": gcs_url},
}
text_message = {
        "type": "text",
        "text": "What is shown in this image?",
    }
message = HumanMessage(content=[text_message, image_message])
```

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/vision_models.py
jzaldi: This is in order for it to work with the chat interface, each turn creates one message right? Maybe I can do so that the first is multimodal?

lkuligin: yes, let's make it multimodal and it should expect only one message, so invocation would look smth like this:

```
response = model.invoke(
        input=[
            HumanMessage(content=[text_message, image_message])
        ]
    )
```

jzaldi: Changed!

lkuligin (approved):

PR #38, opened 2024-02-27
tests: add pytest flags

svidiella:
Add support for flags on pytest and use `--release` flag for integration tests

Support of:
* pytest --release
* pytest --gpu
* pytest --extended
* pytest --long

Flags can be combined, and a single test can be marked  with several flags.
If no flags are provided, the test will be always executed.
If marked with a flag (e.g. @pytest.mark.release) the test will only be executed if the flag is specified in the pytest call.

Marked  all currently not skipped integration tests as `release`.
Marked  `test_llms_safety` tests as extended (as they can be run normally in local, no extra CI setup seems to be needed).


lkuligin:
@movsic take a look, please

lkuligin (approved):

lkuligin:
@efriis fyi, maybe you'll find the approach useful for the core repo

PR #40, opened 2024-02-27
added post-processing for local gemma

lkuligin:


libs/vertexai/langchain_google_vertexai/gemma.py
```diff
@@ -53,6 +53,21 @@ def gemma_messages_to_prompt(history: List[BaseMessage]) -> str:
...
+    pos = response.rfind(pattern)
+    if pos == -1:
+        return response
+    else:
+        text = response[(pos + len(pattern)) :]
+        pos = text.find("<start_of_turn>user\n")
+        if pos > 0:
+            return text[:pos]
+        else:
+            return text
```
holtskinner, line 68:
```suggestion
    text = response[(pos + len(pattern)) :]
    pos = text.find("<start_of_turn>user\n")
    if pos > 0:
        return text[:pos]
    return text
```

holtskinner: else's not needed

libs/vertexai/langchain_google_vertexai/gemma.py
```diff
@@ -139,10 +159,12 @@ async def _agenerate(
         """Top Level call"""
         request = self._get_params(**kwargs)
         request["prompt"] = gemma_messages_to_prompt(messages)
-        output = await self.async_client.predict(
+        text = await self.async_client.predict(
```
holtskinner, line 162: maybe change this back to `output` to be explicit that it's the raw output.

holtskinner (approved):

libs/vertexai/langchain_google_vertexai/gemma.py
lkuligin: ha, true! fixed!

libs/vertexai/langchain_google_vertexai/gemma.py
lkuligin: fixed!

PR #47, opened 2024-02-29
feat: add support for multimodal text embedding

svidiella:


libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -188,30 +191,41 @@ def _get_embeddings_with_retry(
...
+            tasks.append(
+                self.instance["task_executor"].submit(
+                    self.instance["get_embeddings_with_retry"],
+                    contextual_text=text,
+                )
+            )
+        if len(tasks) > 0:
+            wait(tasks)
+        embeddings = [task.result().text_embedding for task in tasks]
+        return embeddings
```
holtskinner, line 214: This only gets the Text embeddings, is that the intended purpose for this method?

libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -375,10 +387,7 @@ def embed_query(self, text: str) -> List[float]:
         Returns:
             Embedding for the text.
         """
-        if self.model_type != GoogleEmbeddingModelType.TEXT:
-            raise NotImplementedError("Not supported for multimodal models")
-        embeddings = self.embed([text], 1, "RETRIEVAL_QUERY")
-        return embeddings[0]
+        return self.embed([text], 1, "RETRIEVAL_QUERY")[0]
```
holtskinner, line 390: I think this needs to only add the "RETRIEVAL_QUERY" if the model is text-only and supports those types.

I don't think those embedding types work with multimodal

holtskinner (approved):

libs/vertexai/langchain_google_vertexai/embeddings.py
svidiella: While text models accept a list of strings for embbeding, multimodal just accepts one each time. I wanted to maintain the user signature which allows to send a list of strings and make them to be embedded in parallel behind.
Adding image and text into the same embedding request will just return the embedding for each of them so I kept them separated for now.

libs/vertexai/langchain_google_vertexai/embeddings.py
svidiella: I wanted to keep the logic the same at this point to take advantage of the code that handle batches and docs and that is currently unified. It's just in the moment to prepare the request and call the LLM that the model type is checked and this is ignored if the model is multimodal. I wanted to maintain the same signature and just allow multimodal models to be used.

lkuligin (approved):

PR #51, opened 2024-03-06
Temp fix

### LINKED ISSUE #49: Temperature parameter is ignored when temperature=0
Hi there, I'm trying to call a gemini-pro model using langchain-google-vertexai. However, as shown in the code below, if temperature=0, the temperature setting does not seem to be reflected.

```python
from langchain_google_vertexai import VertexAI

model1 = VertexAI(model_name="gemini-pro", temperature=0.5)
print(model1._prepare_params())
# {'temperature': 0.5, 'candidate_count': 1, 'stop_sequences': None}

model2 = VertexAI(model_name="gemini-pro", temperature=0)
print(model2._prepare_params())
# {'candidate_count': 1, 'stop_sequences': None}
```

I think the cause may be in the code below:

```python
# https://github.com/langchain-ai/langchain-google/blob/main/libs/vertexai/langchain_google_vertexai/llms.py
# from line 216 of langchain_google_vertexai/llms.py::_VertexAICommon._default_params

    @property
    def _default_params(self) -> Dict[str, Any]:
        ...
        # params is a dictionary of parameters held by the instance, such as temperature.
        for param_name, param_value in params.items():
            default_value = default_params.get(param_name)
            if param_value or default_value:  # if param_value is a value determined to be false, it will be skipped.
                updated_params[param_name] = (
                    param_value if param_value else default_value
                )
        return updated_params
```

I would like to ask why it is implemented this way. I'm a beginner, but if possible, I'm thinking of creating a pull request for the fix. Thanks!

lkuligin:
fixed #49

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2023 LangChain, Inc.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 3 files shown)


eycjur:
@lkuligin 
Thank you for the correction. But i think this fix is ​​not enough. This is because assertCountEqual does not determine the contents of nested dictionaries (I don't know if this is normal behavior for assertCountEqual, but at least it behaves like that)

For example, the following test will pass
```python
from unittest import TestCase

def test_assert_count_equal():
	TestCase().assertCountEqual(
		{
			"stream": False,
			"safety_settings": None,
			"generation_config": {
				"max_output_tokens": 1,
				"temperature": 0,
				"top_k": 10,
				"top_p": 0.5,
				"stop_sequences": None,
			},
		},
		{
			"stream": False,
			"safety_settings": None,
			"generation_config": {
			},
		}
	)
```

And in the case of this pull request, I think it is necessary to modify the line below.

```diff
libs/vertexai/langchain_google_vertexai/llms.py
            default_value = default_params.get(param_name)
+             if param_value is not None or default_value is not None:
-             if param_value or default_value:
                updated_params[param_name] = (
                    param_value if param_value is not None else default_value
                )
        return updated_params
```

Thanks!

lkuligin:
@eycjur it's embarassing for me, but you're right! thanks a lot for spotting it. Fixed.

P.S. Please, feel free to contribute yourself next time, we welcome contributions very much!

eycjur:
@lkuligin 
I've learned a lot from your code, so I'm glad I could help you out!
Your comments are very encouraging. Please let me contribute next time!

PR #64, opened 2024-03-14
Fixed support for custom tool

moiz-stri:


lkuligin:
thank you for your contribution!
please, fix the linter and we can include it in the upcoming release.

lkuligin (approved):

Issue #29, opened 2024-02-26
Adding option to use google credentials for _BaseGoogleGenerativeAI construction


jackklika:
Hi all, I have been using GoogleGenerativeAI in my code. However, in my environment I am using google credentials instead of an api key and the `GoogleGenerativeAI` model validation does not allow me to not specify an api key. I would be interested in contributing this feature but would like some guidance.

My personal workaround has been to override the root validator (mostly copy-paste) as follows:

```
import google.generativeai as genai
from langchain_google_genai import ChatGoogleGenerativeAI
from overrides import overrides
from pydantic.v1 import root_validator

from myapp.clients.google_client import get_creds


# overriding so we can use credentials to build instead of api key
class MyGoogleGenerativeAI(ChatGoogleGenerativeAI):
    @root_validator()
    def validate_environment(cls, values: dict) -> dict:
        #google_api_key = get_from_dict_or_env(
        #    values, "google_api_key", "GOOGLE_API_KEY"
        #)
        #if isinstance(google_api_key, SecretStr):
        #    google_api_key = google_api_key.get_secret_value()

        genai.configure(
            credentials=get_creds(), # <<< type google.auth.credentials.Credentials
            # api_key=google_api_key,
            transport=values.get("transport"),
            client_options=values.get("client_options"),
        )
        if values.get("temperature") is not None and not 0 <= values["temperature"] <= 1:
            raise ValueError("temperature must be in the range [0.0, 1.0]")

        if values.get("top_p") is not None and not 0 <= values["top_p"] <= 1:
            raise ValueError("top_p must be in the range [0.0, 1.0]")

        if values.get("top_k") is not None and values["top_k"] <= 0:
            raise ValueError("top_k must be positive")
        model = values["model"]
        values["client"] = genai.GenerativeModel(model_name=model)
        return values


llm = MyGoogleGenerativeAI(model="gemini-pro")

result = llm.invoke("Write a ballad about LangChain")
print(result.content)
```
My plan to contribute this feature was as follows:
- Add `credentials: Optional[Credentials]` to the `_BaseGoogleGenerativeAI` class
- In `ChatGoogleGenerativeAI` and `GoogleGenerativeAI` root validator, check if `credentials` attribute exists and prioritize using it for `genai.configure()` over the google_api_key logic

However, in my testing the `Credentials` object is not schema-able in pydantic:
```
from google.auth import credentials as ga_credentials
from pydantic import BaseModel
class A(BaseModel):
[...]

PR #78, opened 2024-03-19
Add credentials parameter to GoogleGenerativeAI, ChatGoogleGenerativeAI, and GoogleGenerativeAIEmbedded

jackklika:
Addresses issue #29 -- adds an optional constructor to the generative AI classes to allow passing in a credentials object instead of an api key. If the credentials object is not specified, it falls back to the api key like in current functionality.


libs/genai/langchain_google_genai/llms.py
```diff
@@ -4,6 +4,7 @@
 from typing import Any, Callable, Dict, Iterator, List, Optional, Union
 
 import google.api_core
+import google.cloud.aiplatform
```
lkuligin, line 7: why do you need this import? the genai package doesn't have the dependency on aiplatform.

jackklika: My mistake. I remove this import and fixed formatting on embeddings file.

lkuligin (approved):

PR #88, opened 2024-03-25
code migration to community

lkuligin:


libs/community/langchain_google_community/gcs_file.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def load(self) -> List[Document]:
+        """Load documents."""
+        try:
+            from google.cloud import storage  # type: ignore[attr-defined]
+        except ImportError:
+            raise ImportError(
+                "Could not import google-cloud-storage python package. "
+                "Please install it with `pip install google-cloud-storage`."
+            )
```
svidiella, line 69: Why not adding this as poetry dependencies instead of loading dependencies this way?

lkuligin: that might be the next step, but we need to create separate groups.
we don't want to bring all dependencies together as a requirements for the package.

svidiella (approved):
LGTM 
working on dependencies on next PR

Issue #95, opened 2024-03-26
vertexai: converting pydantic to vertex function breaks with allOf


baskaryan:
pydantic json schema will include an allOf if a nested object has a description. for example

```python
from langchain_core.pydantic_v1 import BaseModel, Field

class Node(BaseModel):
    id: str
    type: str
    
class Relationship(BaseModel):
    source: Node
    target: Node = Field(..., description="foo")


Relationship.schema()
```
generates
```pycon
{
  'title': 'Relationship',
  'type': 'object',
  'properties': {
    'source': {'$ref': '#/definitions/Node'},
    'target': {
      'title': 'Target',
      'description': 'foo',
      'allOf': [{'$ref': '#/definitions/Node'}]
    }
  },
  'required': ['source', 'target'],
  'definitions': {
    'Node': {
      'title': 'Node',
      'type': 'object',
      'properties': {
        'id': {'title': 'Id', 'type': 'string'},
        'type': {'title': 'Type', 'type': 'string'}
      },
     'required': ['id', 'type']
    }
  }
}
```

notice difference between 'source' and 'target'.

When trying to use function-calling, the allOf leads to a protobuf error
```pycon
File ~/anaconda3/lib/python3.11/site-packages/proto/marshal/rules/message.py:36, in MessageRule.to_proto(self, value)
     31 if isinstance(value, dict) and not self.is_map:
     32     # We need to use the wrapper's marshaling to handle
     33     # potentially problematic nested messages.
     34     try:
     35         # Try the fast path first.
---> 36         return self._descriptor(**value)
     37     except TypeError as ex:
     38         # If we have a type error,
     39         # try the slow path in case the error
     40         # was an int64/string issue
     41         return self._wrapper(value)._pb

ValueError: Protocol message Schema has no "allOf" field.
```

using
```
protobuf==4.25.3
google-cloud-aiplatform==1.44.0
```

lkuligin:
@alx13

alx13:
Yep. Unfortunately it's an issue with underlying library:

https://github.com/googleapis/python-aiplatform/blob/cdb8e6afc3791ca5b3c86e516dde2c3f111401f0/google/cloud/aiplatform_v1beta1/types/openapi.py#L63

Pydantic models need to be constructed in a very specific way to avoid unsupported properties with Schema.

Let me check how model can be constructed.

alx13:
It's related with: https://github.com/pydantic/pydantic/issues/3896

[...]

PR #96, opened 2024-03-26
vertexai[patch]: remove allOf from function schema

baskaryan:
patch for #95

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/functions_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/functions_utils.py
@@ -131,7 +131,7 @@ def _get_parameters_from_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
     """
 
     schema = dereference_refs(schema)
-    # TODO: Remove once vertexai api supports allOf elements.
+    # TODO: Remove if vertexai api supports allOf elements in the future.
     schema = _replace_all_ofs(schema)
     model = ParametersSchema.parse_obj(schema)
 
--- a/libs/vertexai/langchain_google_vertexai/functions_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/functions_utils.py
@@ -94,6 +94,31 @@ class ParametersSchema(BaseModel):
         extra = "allow"
 
 
+def _replace_all_ofs(schema: Dict[str, Any]) -> Dict[str, Any]:
+    """Assumes dereferenced schema."""
+    new_schema: Dict[str, Any] = {}
+    for k, v in schema.items():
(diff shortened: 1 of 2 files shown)


baskaryan:
@alx13 @lkuligin

libs/vertexai/langchain_google_vertexai/functions_utils.py
```diff
@@ -94,6 +94,31 @@ class Config:
...
 
 
+def _replace_all_ofs(schema: Dict[str, Any]) -> Dict[str, Any]:
+    """Assumes dereferenced schema."""
+    new_schema: Dict[str, Any] = {}
+    for k, v in schema.items():
+        if isinstance(v, dict) and "allOf" in v:
+            if isinstance(v["allOf"], list) and len(v["allOf"]) == 1:
+                obj = dict(v["allOf"][0])
+                obj["title"] = v.get("title", obj.get("title", ""))
```
alx13, line 104: `title` field is now not supported - we are dropping it

libs/vertexai/langchain_google_vertexai/functions_utils.py
```diff
@@ -94,6 +94,31 @@ class Config:
...
+                obj["description"] = " ".join(
+                    (v.get("description", ""), obj.get("description", ""))
```
alx13, line 106: I'm not sure about joining descriptions.

This code:
```python
class Node(BaseModel):
    """This is a node"""
    id: str
    type: str

class Relationship(BaseModel):
    source: Node
    target: Node = Field(..., description="This is a target node")
```
will produce following definition for target property:
```json
{
  "target": {
    "title": "Target",
    "description": "This is a target node This is a node",
    "type": "object",
    "properties": {
      "id": {
        "title": "Id",
        "type": "string"
      },
      "type": {
        "title": "Type",
        "type": "string"
      }
    },
    "required": ["id", "type"]
  }
}
```

baskaryan: yea not ideal but is there a better generic solution? both descriptions could have important (and distinct) info

alx13: I would consider coalescing i. e. field description should take precedence

```
class Relationship(BaseModel):
    source: Node
    target: Node = Field(..., description="This is a target node")
```

over model description:
```
class Node(BaseModel):
    """This is a node"""
    id: str
    type: str
```

And could you please describe reasoning and behaviour in the function doc string?

alx13: @lkuligin what is your take on that?

PR #106, opened 2024-03-30
added params to invocation methods too

lkuligin:


lkuligin:
@jzaldi PTAL

libs/vertexai/langchain_google_vertexai/vision_models.py
```diff
@@ -38,9 +39,13 @@ class _BaseImageTextModel(BaseModel):
     project: Union[str, None] = Field(default=None)
     """Google cloud project"""
 
-    def _create_model(self) -> ImageTextModel:
-        """Builds the model object from the class attributes."""
-        return ImageTextModel.from_pretrained(model_name=self.model_name)
+    @property
```
jzaldi, line 42: Can't we use here `functools.cached_property` and get rid of the field?

jzaldi (approved):
LGTM with a comment

PR #107, opened 2024-03-31
added image loader with Cloud Vision API

lkuligin:


libs/community/langchain_google_community/vision.py
```diff
@@ -0,0 +1,43 @@
+from typing import Iterator, Optional
+
+from langchain_core.document_loaders import BaseBlobParser
+from langchain_core.document_loaders.blob_loaders import Blob
+from langchain_core.documents import Document
+
+from langchain_google_community._utils import get_client_info
+
+
+class CloudImageLoader(BaseBlobParser):
```
holtskinner, line 10: Can you change this to be `CloudVisionLoader` so that it has the product name in the class?

libs/community/langchain_google_community/vision.py
```diff
@@ -0,0 +1,43 @@
...
+        client_options = None
+        print(project)
+        if project:
+            client_options = {"quota_project_id": project}
+        self._client = vision.ImageAnnotatorClient(
+            client_options=client_options,
+            client_info=get_client_info(module="cloud-vision"),
+        )
+
+    def load(self, gsc_path: str) -> Document:
```
holtskinner, line 28: Can you clarify in the docstring that this only gets the text?

And/or can you add other fields in the `Document` that can get the other data from the Cloud Vision API?

libs/community/langchain_google_community/vision.py
```diff
@@ -0,0 +1,43 @@
...
+            from google.cloud import vision  # type: ignore[attr-defined]
+        except ImportError as e:
+            raise ImportError(
+                "Cannot import google.cloud.vision, please install "
+                "`pip install google-cloud-vision`."
+            ) from e
+        client_options = None
+        print(project)
+        if project:
+            client_options = {"quota_project_id": project}
```
holtskinner, line 22: Can you change this to use the `ClientOptions` class from `google.api_core.client_options`?

libs/community/langchain_google_community/vision.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self, project: Optional[str] = None):
+        try:
+            from google.cloud import vision  # type: ignore[attr-defined]
+        except ImportError as e:
+            raise ImportError(
+                "Cannot import google.cloud.vision, please install "
+                "`pip install google-cloud-vision`."
+            ) from e
+        client_options = None
+        print(project)
```
holtskinner, line 20: Should this `print` be here?

libs/community/langchain_google_community/vision.py
```diff
@@ -0,0 +1,43 @@
...
+
+from langchain_core.document_loaders import BaseBlobParser
+from langchain_core.document_loaders.blob_loaders import Blob
+from langchain_core.documents import Document
+
+from langchain_google_community._utils import get_client_info
+
+
+class CloudImageLoader(BaseBlobParser):
+    def __init__(self, project: Optional[str] = None):
```
holtskinner, line 11: For consistency with other connectors, can this variable be `project_id`

libs/community/langchain_google_community/vision.py
holtskinner, line 28: Change the variable name from `gsc_path` to `gcs_uri`.

Can this also take in a local image?

holtskinner (changes requested):
See comments

libs/community/langchain_google_community/vision.py
lkuligin: actually, for LLM/ChatModels we use `project`, so I'd rather stick to `project` and eventually change everything else to follow the most frequently used integration, wdyt?

libs/community/langchain_google_community/vision.py
lkuligin: for me `CloudImageLoader` sounds more obvious for a user since it loads an image by processing it in the Cloud.

being more easy to grasp for a user is probably more important than having the product name in the class name, wdyt?

libs/community/langchain_google_community/vision.py
lkuligin: removed, thx

libs/community/langchain_google_community/vision.py
lkuligin: good point, added a `source` in the metadata and adjusted docsting

libs/community/langchain_google_community/vision.py
lkuligin: why? the client expect ClientOptions or a dict.

libs/community/langchain_google_community/vision.py
lkuligin: done

wrt a local image, it's a good point, I'll add it as a FR.
I was thinking about using it together with `GCSFileLoader` mostly, but we should think how to use it with a local image too.

libs/community/langchain_google_community/vision.py
holtskinner: I understand what you're saying, but it should probably be very clear what Google Cloud product is being used.

holtskinner (approved):

PR #121, opened 2024-04-08
vertexai[patch]: standardize model params

baskaryan:
Related to https://github.com/langchain-ai/langchain/issues/20085

### DIFF
--- a/libs/vertexai/tests/unit_tests/test_embeddings.py
+++ b/libs/vertexai/tests/unit_tests/test_embeddings.py
@@ -8,6 +8,17 @@ from langchain_google_vertexai import VertexAIEmbeddings
 from langchain_google_vertexai.embeddings import GoogleEmbeddingModelType
 
 
+def test_init() -> None:
+    for embeddings in [
+        VertexAIEmbeddings(
+            model_name="foo",
+            project="test-project",
+        ),
+        VertexAIEmbeddings(model="foo", project="test-project"),
+    ]:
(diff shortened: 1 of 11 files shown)


baskaryan:
mypy doesn't seem to understand the pydantic Field.alias + Config.allow_population_by_field_name typing  :/

libs/vertexai/tests/unit_tests/test_embeddings.py
```diff
@@ -8,6 +8,16 @@
 from langchain_google_vertexai.embeddings import GoogleEmbeddingModelType
 
 
+def test_init() -> None:
+    for embeddings in [
+        VertexAIEmbeddings(
```
baskaryan, line 13: hm seems instantiating vertex embedding model requires setting up vertex creds which isn't great. is there a way to change that in a non-breaking fashion @lkuligin?

lkuligin: ideally it should be supported on the SDK side, I've raised a FR.

lkuligin (approved):

libs/vertexai/tests/unit_tests/test_embeddings.py
lkuligin: so far I'd suggest let's move this to integration tests, wdyt?

baskaryan: moved 👍

PR #123, opened 2024-04-08
update vertex to use tool call attribute on AIMessage

ccurme:


efriis (approved):
Does this also fix some of the streaming issues?

ccurme:
> Does this also fix some of the streaming issues?

not that I know of...

PR #124, opened 2024-04-09
Add Vertex Search Based tool

jzaldi:
- Added `tools.VertexSearchTool` as a tool that returns the summary of the document retrieved.
- Added minimal integration test.


libs/vertexai/langchain_google_vertexai/tools/vertex_search.py
```diff
@@ -0,0 +1,103 @@
...
+        """Runs the tool.
+
+        Args:
+            search_query: The query to run by the agent.
+
+        Returns:
+            The response from the agent.
+        """
+
+        import google.cloud.discoveryengine_v1 as discoveryengine  # type: ignore[import-untyped, unused-ignore]
```
lkuligin, line 59: should it be at langchain_google_community maybe? where we already have an integration for Vertex Search?

PR #125, opened 2024-04-09
Add system_instructions for Gemini models

eliasecchig:


libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -142,24 +142,28 @@ def _convert_to_parts(message: BaseMessage) -> List[Part]:
...
 
     vertex_messages = []
-    raw_system_message = None
+    convert_system_message_to_human_content = None
+    system_instructions = None
     for i, message in enumerate(history):
         if (
             i == 0
             and isinstance(message, SystemMessage)
             and not convert_system_message_to_human
```
lkuligin, line 151: we can probably drop convert_system_message_to_human now, wdyt?

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -142,24 +142,28 @@ def _convert_to_parts(message: BaseMessage) -> List[Part]:
...
-            )
-        elif i == 0 and isinstance(message, SystemMessage):
-            raw_system_message = message
+            if system_instructions is not None:
+                raise ValueError(
+                    "Detected more than one SystemMessage in the list of messages."
+                    "Gemini APIs support the insertion of only SystemMessage."
+                )
+            else:
+                system_instructions = message.content
```
lkuligin, line 159: should we do `_convert_to_parts(message)` here? it looks like system_instructions are multimodal too

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -118,7 +118,7 @@ def _parse_chat_history_gemini(
     history: List[BaseMessage],
     project: Optional[str] = None,
     convert_system_message_to_human: Optional[bool] = False,
-) -> List[Content]:
+) -> dict[str, list[Content] | None | str]:
```
lkuligin, line 121: nits: I'd rather make it a tuple instead of a dict (system part + history)

libs/vertexai/langchain_google_vertexai/chat_models.py
eliasecchig: I thought keeping it for backwards compatibility

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -567,13 +587,17 @@ async def _astream(
...
+        chat_history_result = _parse_chat_history_gemini(
             messages,
             project=self.project,
             convert_system_message_to_human=self.convert_system_message_to_human,
         )
+        history_gemini = chat_history_result["history"]
+        system_instructions = chat_history_result["system_instructions"]
         message = history_gemini.pop()
         with telemetry.tool_context_manager(self._user_agent):
+            if system_instructions is not None:
```
lkuligin, line 596: I don't think we need this if (if it's None, let's reset it to None then otherwise we might cache system instructions from the previous call)

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -118,7 +130,7 @@ def _parse_chat_history_gemini(
     history: List[BaseMessage],
     project: Optional[str] = None,
     convert_system_message_to_human: Optional[bool] = False,
-) -> List[Content]:
+) -> tuple[Any | None, list[Any]]:
```
lkuligin, line 133: nits: can it be tuple[Content | Any, list[Content]] maybe?

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -142,24 +154,30 @@ def _convert_to_parts(message: BaseMessage) -> List[Part]:
...
-                
-To automatically convert the leading SystemMessage to a HumanMessage,
-set  `convert_system_message_to_human` to True. Example:
-
-llm = ChatVertexAI(model_name="gemini-pro", convert_system_message_to_human=True)
-"""
-            )
-        elif i == 0 and isinstance(message, SystemMessage):
-            raw_system_message = message
+            if system_instruction is not None:
```
lkuligin, line 165: this can't be the case since i=0, it's the first message in the list

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -142,24 +154,30 @@ def _convert_to_parts(message: BaseMessage) -> List[Part]:
...
+                system_instruction = Content(
+                    role="user", parts=_convert_to_parts(message)
+                )
+                continue
+        elif (
+            i == 0
+            and isinstance(message, SystemMessage)
+            and convert_system_message_to_human
+        ):
+            convert_system_message_to_human_content = message
```
lkuligin, line 180: should we still throw a warning here and suggest people to switch to gemini-pro-latest?

lkuligin (approved):

libs/vertexai/langchain_google_vertexai/chat_models.py
eliasecchig: ops, yes it should be

libs/vertexai/langchain_google_vertexai/chat_models.py
eliasecchig: I see, you are capturing it below

libs/vertexai/langchain_google_vertexai/chat_models.py
eliasecchig: should be tuple[Content | None, list[Content]] to be correct

PR #135, opened 2024-04-10
added tool config

alx13:


baskaryan (commented):
in this pr or as a followup would be great to bind a tool config in with_structured_output to force tool call

lkuligin:
yes, we've been thinking about it too, need to test it first (whether you need a single step or still two steps)

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -398,25 +397,24 @@ def _generate(
...
                 model_name=self.model_name,
                 safety_settings=safety_settings,
             )
-            with telemetry.tool_context_manager(self._user_agent):
-                chat = self.client.start_chat(history=history_gemini)
 
             # set param to `functions` until core tool/function calling implemented
             raw_tools = params.pop("functions") if "functions" in params else None
             tools = _format_tools_to_vertex_tool(raw_tools) if raw_tools else None
+            tool_config = params.pop("tool_config") if "tool_config" in params else None
```
lkuligin, line 410: btw, @baskaryan @efriis do you see any issues with pasing a proto directly over the Runnable / chain?
or should it always be a dict, and casted to proto here?

efriis: If dicts are supported, that would be preferred on the input/output here. Can have a backup type of a proto, but I think serialization will get weird in e.g. LangServe stream log and stream events.

lkuligin (approved):

PR #148, opened 2024-04-11
standard chat model tests

efriis:
vertex async streaming is broken

probably good to add cases for gemini and palm separately


lkuligin:
closing this for now, enabled on vertexai, and I'll need to take a look what's failing on genai in a separate PR

PR #152, opened 2024-04-12
Fix chat vertexai astream

mathiasburger:
Do not pop the prompt from the conversation history which would pass an empty list of messages to the google client library.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -23,7 +23,7 @@ from typing import (
 
 import proto  # type: ignore[import-untyped]
 from google.cloud.aiplatform_v1beta1.types.content import Part as GapicPart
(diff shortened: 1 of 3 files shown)


libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -722,7 +719,6 @@ async def _astream(
             project=self.project,
             convert_system_message_to_human=self.convert_system_message_to_human,
         )
-        message = history_gemini.pop()
```
mathiasburger, line 725: The variable message is currently unused.

It should not be removed from the history as generate_content_async requires the user message to be present in the contents parameter passed to the client, see https://ai.google.dev/api/python/google/generativeai/GenerativeModel#multi-turn.

lkuligin:
thanks!
it's already fixed here: https://github.com/langchain-ai/langchain-google/pull/151

we'll run a release later today

PR #165, opened 2024-04-18
Fixing multi-part response parsing

alx13:


libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -316,70 +318,88 @@ def _get_client_with_sys_instruction(
...
-                content=content,
-                additional_kwargs=additional_kwargs,
-                tool_call_chunks=tool_call_chunks,
+    tool_calls = []
+    invalid_tool_calls = []
+    tool_call_chunks = []
+
+    for part in response_candidate.content.parts:
+        try:
+            text = part.text
```
lkuligin, line 329: nits: maybe `if hasattr(part, "text") and part.text: ...` instead

alx13: This is a `@property` getter, it would not work

lkuligin (approved):

PR #175, opened 2024-04-22
feat: control all comunity dependencies as poetry groups

svidiella:


libs/community/langchain_google_community/bigquery.py
```diff
@@ -56,7 +56,8 @@ def load(self) -> List[Document]:
         except ImportError as ex:
             raise ImportError(
                 "Could not import google-cloud-bigquery python package. "
-                "Please install it with `pip install google-cloud-bigquery`."
+                "Please, install bigquery dependency group: "
+                "poetry install --with bigquery"
```
lkuligin, line 60: `pip install langchain-google-community[bigquery]` should be better, or am I missing anything?

svidiella: you are right, changing them

svidiella: done!

PR #179, opened 2024-04-23
Add owner, source and size to GoogleDriveLoaders Document's metadata.

rahul-trip:
Description: 
This PR will add 3 new fields to metadata of Documents loaded by GoogleDriveLoader.
A new argument, `load_extended_matadata` is also added to enable these fields. 
Documentation:
Will be added in followup PR (on lagchain repo).

### DIFF
--- a/libs/community/langchain_google_community/drive.py
+++ b/libs/community/langchain_google_community/drive.py
@@ -47,6 +47,108 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
     """The file loader kwargs to use."""
     load_auth: bool = False
     """Whether to load authorization identities."""
+    load_extended_metadata: bool = False
+    """Whether to load extended metadata."""
+
+    def _get_file_size_from_id(self, id: str) -> str:
+        """Fetch the size of the file."""
+        try:
+            import googleapiclient.errors  # type: ignore[import]
+            from googleapiclient.discovery import build  # type: ignore[import]
+        except ImportError as exc:
+            raise ImportError(
+                "You must run "
+                "`pip install --upgrade "
+                "google-api-python-client` "
+                "to load authorization identities."
+            ) from exc
+
+        creds = self._load_credentials()
+        service = build("drive", "v3", credentials=creds)
+        try:
+            file = service.files().get(fileId=id, fields="size").execute()
+            return file["size"]
+        except googleapiclient.errors.HttpError:
+            print(
+                f"insufficientFilePermissions: The user does not have sufficient \
+                permissions to retrieve size for the file with fileId: {id}"
+            )
+            return "unknown"
+        except Exception as exc:
+            print(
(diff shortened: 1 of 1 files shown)


lkuligin:
thanks for the contribution!
for next iterations, we should start adding unit and integration tests to this library.

libs/community/langchain_google_community/drive.py
```diff
@@ -47,6 +47,108 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
...
+        except Exception as exc:
+            print(
+                f"Error occurred while fetching the size for the file with fileId: {id}"
+            )
+            print(f"Error: {exc}")
+            return "unknown"
+
+    def _get_owner_metadata_from_id(self, id: str) -> str:
+        """Fetch the owner of the file."""
+        try:
```
lkuligin, line 86: nits: we should have a local `_import_lib` function to avoid copy-pasting here and there

libs/community/langchain_google_community/drive.py
```diff
@@ -47,6 +47,108 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
...
+                "You must run "
+                "`pip install --upgrade "
+                "google-api-python-client` "
+                "to load authorization identities."
+            ) from exc
+
+        creds = self._load_credentials()
+        service = build("drive", "v3", credentials=creds)
+        try:
+            file = service.files().get(fileId=id, fields="size").execute()
```
lkuligin, line 69: nits: this all can be a small local function (smth like `_load_attribute` and it can take `fields` and `credentials` as parameters)

libs/community/langchain_google_community/drive.py
```diff
@@ -231,6 +337,10 @@ def _load_sheet_from_id(self, id: str) -> List[Document]:
                 }
                 if self.load_auth:
                     metadata["authorized_identities"] = authorized_identities
+                if self.load_extended_metadata:
```
lkuligin, line 340: can we merge this and above section maybe?
`metadata["owner"] = self._get_owner_metadata_from_id(id)`

lkuligin:
@rahul-trip do you think you could address these minor comments please?

lkuligin (approved):

libs/community/langchain_google_community/drive.py
```diff
@@ -47,6 +47,108 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
...
+
+    def _get_file_size_from_id(self, id: str) -> str:
+        """Fetch the size of the file."""
+        try:
+            import googleapiclient.errors  # type: ignore[import]
+            from googleapiclient.discovery import build  # type: ignore[import]
+        except ImportError as exc:
+            raise ImportError(
+                "You must run "
+                "`pip install --upgrade "
```
lkuligin, line 61: please, note we have a group now, so the hint should look like `pip install langchain-google-community[drive]`

PR #180, opened 2024-04-23
Accept request_options in parameters

jzaldi:


libs/genai/langchain_google_genai/llms.py
```diff
@@ -143,6 +144,10 @@ class _BaseGoogleGenerativeAI(BaseModel):
        not return the full n completions if duplicates are generated."""
     max_retries: int = 6
     """The maximum number of retries to make when generating."""
+
+    timeout: Optional[float] = None
```
lkuligin, line 148: should we also support `api_transport` parameter?

jzaldi: This is in the other PR, I openned one per issue but could have been in this one tbf

PR #184, opened 2024-04-24
Enable Model Garden, Vision API and Vertex Standalone RAG APIs tests 

gmogr:


libs/vertexai/tests/integration_tests/test_gemma.py
```diff
@@ -14,16 +14,16 @@
 )
 
 
-@pytest.mark.extended
+@pytest.mark.release
```
lkuligin, line 17: it probably still should be extended

libs/community/tests/integration_tests/test_googlesearch_api.py
```diff
@@ -1,32 +1,41 @@
 """Integration test for Google Search API Wrapper."""
 
-import pytest
+import os
 
 from langchain_google_community.search import GoogleSearchAPIWrapper
 
 
-@pytest.mark.skip(reason="CI/CD not ready.")
 def test_call() -> None:
```
lkuligin, line 8: should we mark it as extended (and add a corresponding flag to pytest conf)?

libs/vertexai/tests/integration_tests/test_model_garden.py
```diff
@@ -100,7 +97,7 @@ async def test_model_garden_agenerate(
 
 @pytest.mark.xfail(reason="CI issue")
```
lkuligin, line 98: should we mark it as extended and drop xfail?

libs/community/tests/integration_tests/test_vision.py
```diff
@@ -1,16 +1,14 @@
...
 
-import pytest
 from langchain_core.document_loaders.blob_loaders import Blob
 from langchain_core.documents import Document
 
 from langchain_google_community import CloudVisionLoader, CloudVisionParser
 
 
-@pytest.mark.skip(reason="CI/CD not ready.")
 def test_parse_image() -> None:
```
lkuligin, line 9: should we mark it as extended?

libs/community/tests/integration_tests/test_googlesearch_api.py
gmogr: done

PR #186, opened 2024-04-26
feat: Vertex Check Grounding API integration

Abhishekbhagwat:
This PR is to integrate the Vertex Standalone API for Check Grounding.
The check grounding API determines how grounded a given piece of text is in a given set of reference texts. The API can generate supporting citations from the reference text to indicate where the given text is supported by the reference texts.
This is implemented as a output parser as Check Grounding is usually done once the generation step of the RAG pipeline is done.

### DIFF
--- a/libs/community/tests/integration_tests/test_check_grounding.py
+++ b/libs/community/tests/integration_tests/test_check_grounding.py
@@ -2,7 +2,7 @@ import os
 from typing import List
 
 import pytest
-from google.cloud import discoveryengine_v1alpha
+from google.cloud import discoveryengine_v1alpha  # type: ignore
 from langchain_core.documents import Document
 
 from langchain_google_community.vertex_check_grounding import (
--- a/libs/community/pyproject.toml
+++ b/libs/community/pyproject.toml
@@ -105,6 +105,7 @@ select = [
 
 [tool.mypy]
 disallow_untyped_defs = "True"
+ignore_missing_imports = "True"
 
 [tool.coverage.run]
 omit = ["tests/*"]
--- a/libs/community/tests/integration_tests/test_rank.py
+++ b/libs/community/tests/integration_tests/test_rank.py
@@ -3,7 +3,7 @@ from typing import List
 from unittest.mock import create_autospec
 
 import pytest
-from google.cloud import discoveryengine_v1alpha
+from google.cloud import discoveryengine_v1alpha  # type: ignore
 from langchain_core.callbacks import CallbackManagerForRetrieverRun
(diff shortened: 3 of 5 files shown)


libs/community/langchain_google_community/vertex_check_grounding.py
```diff
@@ -0,0 +1,222 @@
+from typing import Any, Dict, List, Optional
+
+from google.api_core import exceptions as core_exceptions  # type: ignore
+from google.auth.credentials import Credentials  # type: ignore
+from google.cloud import discoveryengine_v1alpha  # type: ignore
```
lkuligin, line 5: it should not be on the top level, otherwise the library can't be used (since it makes an attempt to import on the module level, at `__init__.py`)

libs/community/langchain_google_community/vertex_check_grounding.py
```diff
@@ -0,0 +1,225 @@
...
+            discoveryengine_v1alpha.GroundingFact(
+                fact_text=doc.page_content,
+                attributes={
+                    key: value
+                    for key, value in doc.metadata.items()
+                    if key not in ["id", "relevance_score"] and value is not None
+                },
+            )
+            for doc in documents
+            if doc.page_content and doc.metadata
```
lkuligin, line 156: why are we ignoring documents with empty metadata?

lkuligin (approved):

libs/community/tests/integration_tests/test_check_grounding.py
```diff
@@ -0,0 +1,116 @@
...
+) -> VertexCheckGroundingWrapper:
+    return VertexCheckGroundingWrapper(
+        project_id=os.environ["PROJECT_ID"],
+        location_id=os.environ.get("REGION", "global"),
+        grounding_config=os.environ.get("GROUNDING_CONFIG", "default_grounding_config"),
+        client=grounded_generation_service_client,
+    )
+
+
+def test_integration_parse(
```
lkuligin, line 95: should any additional APIs be enabled on the project where tests are running?

lkuligin: I mean, should any GCP APIs be enabled / whitelisted for the GCP project where this test is running at?

Abhishekbhagwat: Yes, we would need to whitelist Discovery Engine API if it has not been whitelisted for this to run.

lkuligin: we need to mark it as `extended` then

PR #205, opened 2024-04-30
feat: Added features for May 2024 Embeddings Models

holtskinner:
https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings#latest_models

- Add Output Dimensionality
- Add additional embeddings types


lkuligin:
could you extend an integration test, please?

libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -109,6 +109,9 @@ def __init__(
         self.instance[
             "embeddings_task_type_supported"
         ] = not self.client._endpoint_name.endswith("/textembedding-gecko@001")
+        self.instance["dimensionality_supported"] = self.client._endpoint_name.endswith(
```
lkuligin, line 112: could you move this logic to _utils, please, so that we have all the logic determening model's behaviour at least at one place (and we should introduct model family or smth like this).

lkuligin: I've added ModelFamily notion, let's use it, please.

libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -215,7 +222,10 @@ def _get_multimodal_embeddings_with_retry(
         return embeddings
 
     def _get_text_embeddings_with_retry(
-        self, texts: List[str], embeddings_type: Optional[str] = None
+        self,
+        texts: List[str],
+        embeddings_type: Optional[str] = None,
+        dimensions: Optional[int] = None,
```
lkuligin, line 228: can we stick to the argument's naming the same as Vertex uses, please? i.e., `output_dimensionality` instead of `dimensions`?

lkuligin (approved):

PR #207, opened 2024-05-01
refactored, got rid of start_chat

lkuligin:


libs/genai/langchain_google_genai/_genai_extension.py
```diff
@@ -225,14 +225,57 @@ def build_semantic_retriever() -> genai.RetrieverServiceClient:
...
+
+
+def build_generative_async_service(
+    credentials: Optional[credentials.Credentials],
+    api_key: Optional[str] = None,
+    client_options: Optional[Dict[str, Any]] = None,
+    client_info: Optional[gapic_v1.client_info.ClientInfo] = None,
+    transport: Optional[str] = None,
+) -> genai.GenerativeServiceAsyncClient:
+    formatted_client_options = {"api_endpoint": _config.api_endpoint}
```
alx13, line 262: This code is repeated with previous function, can we merge it to like "prepeare_params" func and use it on chat_models.py:627

alx13 (approved):

libs/genai/langchain_google_genai/_genai_extension.py
lkuligin: fixed, thx

PR #217, opened 2024-05-12
Add optional id_field to VertexRank

shenghann:
There are valid use cases to retain the `id` field in the original retrieved documents in the documents that are reranked by `VertexRank`. This is especially important for use cases where traceability of documents throughout the chain is important.
 
Current assigned `id` based on an enumerated incrementing index works for cases where `id`s do not need to be tracked but isn't universally applicable.


libs/community/langchain_google_community/vertex_rank.py
```diff
@@ -114,7 +117,11 @@ def _rerank_documents(
 
         records = [
             discoveryengine_v1alpha.RankingRecord(
-                id=str(idx),
+                id=(
+                    doc.metadata.get(self.id_field)
```
lkuligin, line 121: should we maybe raise an error instead if the `id_field` is not in metadata?
or should we at least either always use enumerator OR id from metadata (i.e., if at least in one docs it's missing, let's use enumerator and raise a warning)?

shenghann: good idea. will add a check if `id_field` is missing: raise a `KeyError`

shenghann: added!

lkuligin:
@shenghann linter is red

libs/community/langchain_google_community/vertex_rank.py
```diff
@@ -112,9 +115,18 @@ def _rerank_documents(
         """
         from google.cloud import discoveryengine_v1alpha  # type: ignore
 
+        if self.id_field:
+            for doc in documents:
+                if self.id_field not in doc.metadata:
+                    raise KeyError(f"id_field '{self.id_field}' not found in document metadata.")
```
lkuligin, line 121: nits: we could just modify below `doc.metadata[self.id_field] if ... else ...`, wrap the block below with `try...except` and catch the KeyError, it would be a little bit nicer :)

lkuligin (approved):

libs/community/langchain_google_community/vertex_rank.py
shenghann: done. raising a warning when KeyError occurs

lkuligin:
@shenghann you can run linter locally with `make format` and `make lint` commands
https://github.com/langchain-ai/langchain-google/blob/main/libs/community/Makefile

lkuligin (approved):

lkuligin (approved):

PR #227, opened 2024-05-16
Add media support to vertex chat_models

wafle:
Fixes https://github.com/langchain-ai/langchain/issues/20738.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -82,10 +82,12 @@ from vertexai.preview.language_models import (
 )
 
 from google.cloud.aiplatform_v1beta1.types import (
+    Blob,
     Candidate,
     Part,
     HarmCategory,
     Content,
+    FileData,
     FunctionCall,
     FunctionResponse,
     GenerateContentRequest,
@@ -93,6 +95,7 @@ from google.cloud.aiplatform_v1beta1.types import (
     SafetySetting,
     Tool as GapicTool,
     ToolConfig as GapicToolConfig,
+    VideoMetadata,
 )
 from langchain_google_vertexai._base import _VertexAICommon, GoogleModelFamily
 from langchain_google_vertexai._image_utils import ImageBytesLoader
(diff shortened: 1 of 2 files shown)


libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -194,7 +195,19 @@ def _convert_to_prompt(part: Union[str, Dict]) -> Part:
             path = part["image_url"]["url"]
             return ImageBytesLoader(project=project).load_gapic_part(path)
 
-        raise ValueError("Only text and image_url types are supported!")
+        # Handle media type like LangChain.js
+        # https://github.com/langchain-ai/langchainjs/blob/e536593e2585f1dd7b0afc187de4d07cb40689ba/libs/langchain-google-common/src/utils/gemini.ts#L93-L106
+        if part["type"] == "media":
+            if "mime_type" not in part:
```
wafle, line 201: I went with snake_case for mime_type and file_uri contrary to what's happening in langchain.js as it seemed more pythonic. Do you think that's the right call?

lkuligin: can we construct the proto Part directly? why do we need a VertexPart?
and it's probably a good idea to allow to pass gcs_uri as file_data and video_metadata too.

libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -140,6 +142,62 @@ def test_multimodal() -> None:
...
+    assert isinstance(output.content, str)
+
+
+@pytest.mark.release
+@pytest.mark.parametrize("file_uri,mime_type", multimodal_inputs)
+def test_multimodal_media_inline_base64(file_uri, mime_type) -> None:
+    llm = ChatVertexAI(model_name="gemini-1.5-pro-preview-0514")
+    storage_client = storage.Client()
+    blob = storage.Blob.from_string(file_uri, client=storage_client)
+    media_base64 = base64.b64encode(blob.download_as_bytes()).decode()
```
lkuligin, line 186: could you include a test that passes a gcs_uri, please?

libs/vertexai/langchain_google_vertexai/chat_models.py
wafle: >can we construct the proto Part directly?

Done.

>video_metadata

Done.

>pass gcs_uri

file_uri expects a GCS uri, e.g. "gs://bucket/some_video.mp4". Do you mean we should also support http uri-s here?

libs/vertexai/tests/integration_tests/test_chat_models.py
wafle: See test_multimodal_media_file_uri.

libs/vertexai/langchain_google_vertexai/chat_models.py
IA-Programming: so there's no way to send the video into a base64 uri?, instead of an GCS uri?

Issue #144, opened 2024-04-11
Vertex AI [feature]: Support JSON Mode for Gemini


afirstenberg:
See documentation for `responseMimeType` at https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini#request_body

See https://github.com/langchain-ai/langchainjs/issues/5071

Ben81828:
Is it possible to pass in our defined schema and specify the use of json_mode, like in this example?
https://python.langchain.com/v0.1/docs/modules/model_io/chat/structured_output/

```python
structured_llm = model.with_structured_output(Joke, method="json_mode")
```

PR #228, opened 2024-05-17
add response_mime_type for gemini models (allow JSON-mode output, optionally)

nicsuzor:
Minor change to optionally add response_mime_type to generation_config. Note this is a preview feature in the Vertex API available for gemini-1.5-pro: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/gemini#parameters
 
Fulfills #144.


lkuligin:
Thanks for your contribution!

`response_mime_type` was already supported and passed as part as `kwargs`:
https://github.com/langchain-ai/langchain-google/blob/d4dcff348e02751605d487b5c1979c7369437636/libs/vertexai/langchain_google_vertexai/chat_models.py#L122

if you feel like we should also allow to setup its default value during model initialization, I'd suggest to  redefine the `_default_params` property (by adding  `response_mime_type` to super()._default_params).

So that similar things are defined at similar place, and the logic not being spread all over the code base. WDYT?

nicsuzor:
> if you feel like we should also allow to setup its default value during model initialization, I'd suggest to redefine the `_default_params` property (by adding `response_mime_type` to super()._default_params).
> 
> So that similar things are defined at similar place, and the logic not being spread all over the code base. WDYT?

Sure thing, I've updated that now.

libs/vertexai/langchain_google_vertexai/_base.py
```diff
@@ -189,6 +196,7 @@ def _default_params(self) -> Dict[str, Any]:
             "temperature": self.temperature,
             "max_output_tokens": self.max_output_tokens,
             "candidate_count": self.n,
+            "response_mime_type": self.response_mime_type,
```
lkuligin, line 199: I don't think it should be there, since `response_mime_type` is not support for embeddings.

lkuligin: It should be rather a redefined property on a chat model that uses super()._default_params

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -709,11 +709,14 @@ def _prepare_request_gemini(
         functions: Optional[List[_FunctionDeclarationLike]] = None,
         tool_config: Optional[Union[_ToolConfigDict, ToolConfig]] = None,
         safety_settings: Optional[SafetySettingsType] = None,
+        response_mime_type: Optional[str] = None,
```
lkuligin, line 712: I don't think we need this line anymore, do we? it will be passed as a part of `kwargs`.

lkuligin (changes requested):

lkuligin:
@nicsuzor could you finalize it, please?

libs/vertexai/langchain_google_vertexai/_base.py
nicsuzor: removed

libs/vertexai/langchain_google_vertexai/chat_models.py
nicsuzor: removed

nicsuzor:
> @nicsuzor could you finalize it, please?

Yup, edited accordingly.

lkuligin (approved):

PR #237, opened 2024-05-20
genai and vertexai: implement ls_params

ccurme:
This implements a set of [standard params](https://github.com/langchain-ai/langchain/blob/2316635adda5db75a9773a995f790079fdfbc647/libs/core/langchain_core/language_models/chat_models.py#L66) that are referenced in Langsmith. It will allow Google GenAI and VertexAI models to support certain Langsmith features.


efriis (approved):

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -614,6 +615,23 @@ def validate_environment(cls, values: Dict) -> Dict:
     def _is_gemini_advanced(self) -> bool:
         return self.model_family == GoogleModelFamily.GEMINI_ADVANCED
 
+    def _get_ls_params(
+        self, stop: Optional[List[str]] = None, **kwargs: Any
+    ) -> LangSmithParams:
+        """Get standard params for tracing."""
+        params = self._get_invocation_params(stop=stop, **kwargs)
```
lkuligin, line 622: maybe use `self._prepare_params` instead to respect kwargs (there's some mapping of arg names happening there)? Or probably we should rather redefine `_getinvocation_params` method for Vertex integrations :).

https://github.com/langchain-ai/langchain-google/blob/2f7be6f1d6b80fab0b2059a1265c0114120013fe/libs/vertexai/langchain_google_vertexai/_base.py#L220

ccurme: thanks, made this change.

PR #244, opened 2024-05-21
require that async client be initialized inside a running event loop

vbarda:
This change fixes the issue with running `ChatGoogleGenerativeAI` inside a `ThreadPoolExecutor` and requires initializing the model inside an event loop to enable async client

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import base64
 import json
 import logging
@@ -558,6 +559,14 @@ def _response_to_result(
     return ChatResult(generations=generations, llm_output=llm_output)
 
 
+def _is_event_loop_running():
+    try:
+        asyncio.get_running_loop()
+        return True
+    except RuntimeError:
+        return False
+
+
 class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
     """`Google Generative AI` Chat models API.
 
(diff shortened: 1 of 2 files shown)


libs/genai/tests/unit_tests/test_chat_models.py
```diff
@@ -50,6 +52,25 @@ def test_integration_initialization() -> None:
     )
 
 
+def test_initialization_inside_threadpool() -> None:
+    with ThreadPoolExecutor() as executor:
```
nfcampos, line 56:
```suggestion
    # new threads don't have a running event loop, thread pool executor easiest way to create one
    with ThreadPoolExecutor() as executor:
```

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -639,13 +648,15 @@ def validate_environment(cls, values: Dict) -> Dict:
...
-        values["async_client"] = genaix.build_generative_async_service(
-            credentials=values.get("credentials"),
-            api_key=google_api_key,
-            client_info=client_info,
-            client_options=values.get("client_options"),
-            transport=transport,
-        )
+
+        if _is_event_loop_running():
+            values["async_client"] = genaix.build_generative_async_service(
```
efriis, line 653: can we add a comment that this is a limitation of `genaix`?

nfcampos (approved):

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -639,13 +648,15 @@ def validate_environment(cls, values: Dict) -> Dict:
...
+                credentials=values.get("credentials"),
+                api_key=google_api_key,
+                client_info=client_info,
+                client_options=values.get("client_options"),
+                transport=transport,
+            )
```
efriis, line 659: maybe good to add an `else` case setting `values["async_client"] = values.get("async_client", None)` just to be explicit (pydantic does funny things sometimes)

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -733,6 +744,10 @@ async def _agenerate(
             tool_config=tool_config,
             generation_config=generation_config,
         )
+
+        if not self.async_client:
```
efriis, line 748: move above `_prepare_request`

efriis (commented):
is there anything we can do to make this usable in a setting without an eventloop? Or does it need to be fixed in the upstream sdk?

vbarda:
> is there anything we can do to make this usable in a setting without an eventloop? Or does it need to be fixed in the upstream sdk?

I believe it would have to be fixed in the SDK

cc @lkuligin

PR #256, opened 2024-05-23
add telemtery to vectorstore

jzaldi:


libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
```diff
@@ -176,3 +180,11 @@ def _get_deployed_index_id(self) -> str:
             f"deployed on endpoint "
             f"{self._endpoint.display_name}."
         )
+    
+
+    def _get_telemetry_user_agent(self) -> str:
```
lkuligin, line 185: do we really need this as a separate function? it's used only once

jzaldi: Maybe we don't wanted to be consistent with the other integrations, let me remove it and we can recrate it if we need it later on

lkuligin (approved):

PR #257, opened 2024-05-23
Fix community tests

gmogr:


lkuligin (approved):

lkuligin (approved):

libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -524,7 +522,6 @@ class MyModel(BaseModel):
 
 
 @pytest.mark.release
-@pytest.mark.xfail(reason="flaky")
```
lkuligin, line 527: I marked this test as flaky yesterday, let's keep it as it is (this is the reason the pipeline failing).

PR #264, opened 2024-05-30
Add fs vectorstore

eliasecchig:


libs/community/langchain_google_community/bq_storage_vectorstores/_base.py
```diff
@@ -0,0 +1,520 @@
...
+_vector_table_lock = Lock()  # process-wide BigQueryVectorSearch table lock
+
+# Constants for index creation
+MIN_INDEX_ROWS = 5
+INDEX_CHECK_INTERVAL = timedelta(seconds=60)
+USER_AGENT_PREFIX = "FeatureStore"
+
+
+class BaseBigQueryVectorStore(VectorStore, BaseModel):
+    embedding: Any
```
lkuligin, line 33: nits: can a type here be `Embedding`?

libs/community/langchain_google_community/bq_storage_vectorstores/_base.py
```diff
@@ -0,0 +1,520 @@
...
+    doc_id_field: str = "doc_id"
+    credentials: Optional[Any] = None
+    embedding_dimension: Optional[int] = None
+    _extra_fields: Union[Dict[str, str], None] = None
+    _table_schema: Any = None
+
+    def sync(self) -> None:
+        raise NotImplementedError()
+
+    def get_documents(
```
lkuligin, line 49: let's mark this class an abstract one, please, and mark method as an abstract one too.

libs/community/langchain_google_community/bigquery_vector_search.py
```diff
@@ -89,6 +89,10 @@ def __init__(
             credentials (Credentials, optional): Custom Google Cloud credentials
                 to use. Defaults to None.
         """
+        logging.warning(
```
lkuligin, line 92: LC has a deprecated flag for this purposes, please use it.

libs/community/langchain_google_community/bq_storage_vectorstores/featurestore.py
```diff
@@ -0,0 +1,527 @@
...
+                        "DNS certificates are being propagated,"
+                        " waiting for 10 seconds.  "
+                    )
+                    time.sleep(10)
+
+                else:
+                    raise
+
+    def sync(self) -> None:
+        """Sync the data from the Big Query source into the Executor source"""
```
ashleyxuu, line 166: nit: BigQuery

libs/community/langchain_google_community/bq_storage_vectorstores/featurestore.py
```diff
@@ -0,0 +1,527 @@
...
+                (ie [{"url": "www.myurl1.com", "title": "title1"},
+                {"url": "www.myurl2.com", "title": "title2"}])
+        Returns:
+            List of ids from adding the texts into the vectorstore.
+        """
+        vs_obj = VertexFSVectorStore(embedding=embedding, **kwargs)
+        vs_obj.add_texts(texts, metadatas)
+        return vs_obj
+
+    def get_big_query_vector_store(self, **kwargs: Any) -> Any:
```
ashleyxuu, line 519: Do we want to rename it to`get_bigquery_vector_store`?

libs/community/langchain_google_community/bq_storage_vectorstores/_base.py
```diff
@@ -0,0 +1,528 @@
...
+        )
+        if self.embedding_dimension is None:
+            self.embedding_dimension = len(self.embedding.embed_query("test"))
+        self._full_table_id = (
+            f"{self.project_id}." f"{self.dataset_name}." f"{self.table_name}"
+        )
+        self._initialize_bq_table()
+        self._validate_bq_table()
+        self._logger.info(
+            f"BigQuery xtable {self._full_table_id} "
```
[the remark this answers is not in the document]
eliasecchig: fixed, thanks!

libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py
```diff
@@ -0,0 +1,489 @@
...
+                ON `{self._full_table_id}`
+                ({self.text_embedding_field})
+                OPTIONS(distance_type="{self.distance_type}", index_type="IVF")
+            """
+            self._bq_client.query(sql).result()  # type: ignore[union-attr]
+            self._have_index = True
+        except ClientError as ex:
+            self._logger.debug("Vector index creation failed (%s).", ex.args[0])
+        finally:
+            self._creating_index = False
```
jinmmin, line 168: I notice that the FeatureStore part has  the functionality to wait for things to be ready. 
BQ index is created partially asynchronously. The return of the DDL does not indicate that the index is finished. Users actually need to check the INFORMATION SCHEMA or run a query to know that.

I wonder if we should have the option for users to wait for BQ index to be ready; or document the behavior in the comment or the langchain document.

lspataroG: @jinmmin This is a good point. For BQ we retained the functionality of the former BigQueryVectorSearch class.
Having said that, I think the difference with FS is that in BQ it is still possible to query on the full set of data even if the index is not ready. On the contrary, for FS you need the sync process to finish in order to be able to query on the whole dataset.

lkuligin (approved):

PR #302, opened 2024-06-16
vertex: fix test

ccurme:
Test has failed last two days with:

> google.api_core.exceptions.InvalidArgument: 400 Please ensure that function response turn comes immediately after a function call turn. And the number of function response parts should be equal to number of function call parts of the function call turn.

There should be one ToolMessage per tool call.


efriis (approved):
Just confirming my understanding - this was just a bug in the test because we were passing one of the tool messages twice

lkuligin (approved):

ccurme:
> Just confirming my understanding - this was just a bug in the test because we were passing one of the tool messages twice

Yes, my understanding is that previously we were making tool calls 1, 2, 3 and passing in tool messages 3, 1, 2, 3.

PR #317, opened 2024-06-19
community[patch], infra: fix pydantic import and add import ci

baskaryan:


libs/vertexai/tests/integration_tests/test_tools.py
```diff
@@ -45,231 +34,3 @@ def parse_result(
 
     def parse(self, text: str) -> Union[AgentAction, AgentFinish]:
         raise ValueError("Can only parse messages")
-
-
-@pytest.mark.xfail(reason="investigating")
```
lkuligin, line 50: why do you think we should remve these tests?

baskaryan: all tool calling abilities are now tested by the standard tests so they're redundant

lkuligin (approved):

PR #319, opened 2024-06-20
Expose SafetySetting enum as a type

kardiff18:
Add SafetySetting type as part of enums, so that the `method` parameter can be properly set. Otherwise, the way to do it now involves importing from VertexAI natively which is a bit clunky. 

```
from google.cloud.aiplatform_v1beta1.types import SafetySetting
from langchain_google_vertexai import HarmBlockThreshold, HarmCategory


safety_settings = [
    SafetySetting(
        category=HarmCategory.HARM_CATEGORY_DANGEROUS_CONTENT,
        threshold=HarmBlockThreshold.BLOCK_LOW_AND_ABOVE,
        method=SafetySetting.HarmBlockMethod.SEVERITY
    )
]
```


lkuligin (approved):

kardiff18:
@lkuligin sorry for the new commits/messages. I ended up adding an additional field to return so that users can see the severity in the response if they set debug to True while running Langchain. Otherwise, it's not returned so it gives off the impression that it isn't being used.

PR #320, opened 2024-06-21
Bump google-generativeai version

jayfish0:
google-generativeai version from "^0.5.2" to "^0.6.0"


jayfish0:
bumping the version would allow some post-processing method possible with google-generativeai library that only exist in the 0.6.0 version, for example https://ai.google.dev/api/python/google/generativeai/types/GenerateContentResponse#to_dict, this method is the only way reliable way to serialize GenerateContentResponse

lkuligin:
closing this one for now, in the future - please, don't forget to update the lock file too.

PR #323, opened 2024-06-21
fix pydantic v1 errors, remove bigtable type

eliasecchig:


lkuligin (approved):

libs/community/langchain_google_community/bq_storage_vectorstores/featurestore.py
```diff
@@ -59,16 +59,10 @@ class VertexFSVectorStore(BaseBigQueryVectorStore):
             store. Defaults to the dataset name.
         online_store_location (str, optional): Location of the online store. Default
             to "location" parameter.
-        online_store_type (Literal["bigtable", "optimized"]): Type of online store.
-            Defaults to "optimized".
+        online_store_type (Literal["optimized"]): Type of online store. Currently only
```
lkuligin, line 62: should we drop it completely then?

PR #331, opened 2024-06-26
feat: Add Vertex Vector Search Datapoints deletion, add autogenerated datapoint IDs to document metadata during ingestion

Abhishekbhagwat:
Support Vertex Vector Search Datapoints deletion via the `vector_store.delete()` method


libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
```diff
@@ -192,6 +192,10 @@ def add_texts(
                 f"{len(metadatas)} != {len(texts)}"
             )
 
+        # Add generated IDs to metadata
```
lkuligin, line 195: I'm not sure modifying things in-place is a good idea. Can we just pass doc_ids from outside maybe?

lkuligin:
@Abhishekbhagwat are you planning to finalize this one?

PR #333, opened 2024-06-27
[Breaking] change the behavior of ChatVertexAI.with_structured_output when specifying a dict

kiarina:
The behavior of `ChatVertexAI.with_structured_output` when specifying a `dict` was different from `ChatOpenAI` and `ChatAnthropicVertex`, so it has been corrected.

- Modified to accept not only the `["name", "description", "parameters"]` format for the `dict` but also `["title", "description", "properties"]`.
- When a `dict` is specified as the schema, it now extracts properties from the output.

This is considered breaking as it changes existing functionality.
If this change is unacceptable, please feel free to close it.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -52,13 +52,14 @@ from langchain_core.messages import (
 from langchain_core.messages.ai import UsageMetadata
 from langchain_core.output_parsers.base import OutputParserLike
 from langchain_core.output_parsers.openai_tools import (
-    JsonOutputToolsParser,
+    JsonOutputKeyToolsParser,
     PydanticToolsParser,
 )
 from langchain_core.output_parsers.openai_tools import parse_tool_calls
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
 from langchain_core.pydantic_v1 import BaseModel, root_validator, Field
 from langchain_core.runnables import Runnable, RunnablePassthrough
+from langchain_core.utils.function_calling import convert_to_openai_function
 from vertexai.generative_models import (  # type: ignore
     Tool as VertexTool,
 )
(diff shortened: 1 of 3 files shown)


lkuligin (approved):

lkuligin:
asking @efriis for the second opinion :), but the change looks reasonable to me.
thanks for the PR!

efriis (commented):
I think this will also break cases where a protobuf is passed in, which might be a useful feature (to match what vertex api provides). Will take a closer look when I'm back from vacation next week, and @baskaryan might have some takes before then!

The common abstraction currently includes pydantic BaseModel

The question seems to be whether we want to have the common abstraction also include:
1. Whatever the provider API supports by default
2. JSON Schema (OpenAI/Anthropic)

In general I'm more supportive of 1 (and closing this), and @baskaryan may have different input!

baskaryan (commented):
its a good change but technically a breaking one, and dont think we can call it a bugfix. if we land we'll probably want to accompany with a minor version bump

baskaryan:
> I think this will also break cases where a protobuf is passed in, which might be a useful feature (to match what vertex api provides). Will take a closer look when I'm back from vacation next week, and @baskaryan might have some takes before then!
> 
> The common abstraction currently includes pydantic BaseModel
> 
> The question seems to be whether we want to have the common abstraction also include:
> 
> 1. Whatever the provider API supports by default
> 2. JSON Schema (OpenAI/Anthropic)
> 
> In general I'm more supportive of 1 (and closing this), and @baskaryan may have different input!

re: breaking support for protobuf schemas, agree we should avoid this. could we either explicitly check the type of the schema before calling convert_to_openai_function, or else try/except that conversion. and when the schema cannot be converted to openai format we keep current behavior

kiarina:
> re: breaking support for protobuf schemas, agree we should avoid this. could we either explicitly check the type of the schema before calling convert_to_openai_function, or else try/except that conversion. and when the schema cannot be converted to openai format we keep current behavior

I've made it so that the behavior changes only when a dict containing title, description, and properties is passed. How does this look?

baskaryan (approved):

PR #341, opened 2024-06-28
vertexai[patch], genai[patch]: use tool name explicitly in with struc…

baskaryan:
…tured output tool binding

part of https://github.com/langchain-ai/langchain/issues/23644


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -959,7 +959,7 @@ def with_structured_output(
             )
         else:
             parser = JsonOutputToolsParser()
-        llm = self.bind_tools([schema], tool_choice=False)
```
baskaryan, line 962: seems i need to add a check for gemini 1.5

lkuligin (approved):

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1006,3 +1007,14 @@ def bind_tools(
             ]
             tool_config = _tool_choice_to_tool_config(tool_choice, all_names)
         return self.bind(tools=genai_tools, tool_config=tool_config, **kwargs)
+
+    @property
+    def _supports_tool_choice(self) -> bool:
+        return "gemini-1.5" in self.model
```
lkuligin, line 1013: nits: there's a special enum for that:
https://github.com/langchain-ai/langchain-google/blob/e7b45726adc15e17f75f018f4542ec11aecdaeb3/libs/vertexai/langchain_google_vertexai/_utils.py#L120

ccurme (approved):

libs/genai/langchain_google_genai/chat_models.py
baskaryan: but thats in vertexai

lkuligin (approved):

PR #350, opened 2024-07-03
[ChatVertexAI] Get `api_transport` from  initializer when not provided

jzaldi:
- Pulls the `api_transport` parameter from the default initializer if not provided


jzaldi:
Maybe we can do something more general because we are overriding vertexai initialization right (other than for project)

lkuligin (approved):

PR #352, opened 2024-07-04
Add Context catching to VertexAI Models

jzaldi:
- Add a parameter `cached_content` to `ChatVertexAI` to allow reuse of cached content.
- Add function `create_context_cache` to create a model + messages cache.

Implementation as per:

https://cloud.google.com/vertex-ai/generative-ai/docs/context-cache/context-cache-overview


libs/vertexai/langchain_google_vertexai/_content_catching.py
```diff
@@ -0,0 +1,75 @@
...
+)
+from langchain_google_vertexai.functions_utils import (
+    _format_to_gapic_tool,
+    _format_tool_config,
+    _ToolConfigDict,
+    _ToolsType,
+)
+
+
+def create_context_cache(
```
lkuligin, line 19: since it's in a private module, afaik, it won't make it to the public python docs:
https://api.python.langchain.com/en/latest/google_vertexai_api_reference.html

should we move it elsewhere?

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -991,6 +991,11 @@ class Joke(BaseModel):
         The format of the dictionary should follow Open API schema.
     """
 
+    cached_content: Optional[str] = None
```
lkuligin, line 994: should we add it is a kwarg to the `_generate` method too?

and maybe add a note to docsting on how to use caching.

libs/vertexai/langchain_google_vertexai/_content_catching.py
jzaldi: You are right! I can make it public. I didn't want to put it in `ChatModels` because that file is getting a little bit big. What do you think?

libs/vertexai/langchain_google_vertexai/chat_models.py
jzaldi: I'll do it!

libs/vertexai/langchain_google_vertexai/_content_catching.py
lkuligin: maybe it's time to introduce public utils :).

lkuligin (approved):

PR #318, opened 2024-06-20
Multithreaded GCSDocumentStorage uploads and downloads

atognolag:
The GCSDocumentStorage now leverages:
1. The google.cloud.storage.transfer_manager for multithreaded uploads and downloads of Blobs
2. The google.cloud.storage.client.batch context manager to batch delete requests

A couple tests were added:
1. Validation of user provided threading configurations
2. Testing threaded and unthreaded operations specifically
3. Tests now upload documents with characters from many different languages

Known limitations:
1. The unthreaded uploads generate plain text documents in GCS, while the multithreaded one generates also plain text documents that GCS reports as application/octet-stream

PR #361, opened 2024-07-09
Make GCSDocumentStorage.mset 10x faster and add default retry

TupleType:
Currently GCSDocumentStorage.mset uploads files at a rate of about 200 per minute. That is very slow for large scale production projects so I refactored it to use transfer manager to get to a speed of about 2000 files per minute.
In addition currently this method fails on backend error which occur from time to time as mentioned in the GCS documentation. To prevent failures I've enabled the default retry mechanism.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vectorstores/document_storage.py
+++ b/libs/vertexai/langchain_google_vertexai/vectorstores/document_storage.py
@@ -1,7 +1,11 @@
 from __future__ import annotations
 
 import json
+from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import TYPE_CHECKING, Any, Dict, Iterator, List, Optional, Sequence, Tuple
+from google.cloud.storage import transfer_manager
+from google.cloud.storage.retry import DEFAULT_RETRY
 
 from google.cloud import storage  # type: ignore[attr-defined, unused-ignore]
 from langchain_core.documents import Document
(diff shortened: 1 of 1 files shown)


libs/vertexai/langchain_google_vertexai/vectorstores/document_storage.py
```diff
@@ -39,16 +43,26 @@ def mset(self, key_value_pairs: Sequence[Tuple[str, Document]]) -> None:
...
+                with open(Path(tmp_folder) / key, "w") as f:
+                    json.dump(value.dict(), f)
+
+            transfer_manager.upload_many_from_filenames(
+                self._bucket,
+                [item[0] for item in key_value_pairs],
+                blob_name_prefix=f"{self._prefix}/",
+                source_directory=tmp_folder,
+                upload_kwargs={"retry": DEFAULT_RETRY},
+            )
```
jzaldi, line 57: Looks pretty good! The only thing I'm not quite sure about is the use of the disk to stage the files, have you tried `upload_many` instead?
https://cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.transfer_manager

TupleType: I've looked at it but I don't think it's the right choice because often uploading documents includes large amount of files which will not fit in memory

jzaldi:
Actually, this is duplicated with #318

TupleType:
> Actually, this is duplicated with #318

My bad. You can close this if you think it's not needed

PR #363, opened 2024-07-12
Fix optional fields in Pydantic models

orkhank:
## Problem

[Pylance](https://marketplace.visualstudio.com/items?itemName=ms-python.vscode-pylance) is flagging initialization of some [Pydantic](https://docs.pydantic.dev/latest/) objects (ie. objects of classes extending [`BaseModel`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel)) as having missing arguments when optional arguments are not provided. Although the code still runs as it should, this type of error is a potential source of frustration for VS Code users.

### Example

For example, according to the doc string for the [`GoogleGenerativeAIEmbeddings`](https://api.python.langchain.com/en/latest/embeddings/langchain_google_genai.embeddings.GoogleGenerativeAIEmbeddings.html) class, it can be used in the following way:

```python

embeddings = GoogleGenerativeAIEmbeddings(model="models/embedding-001")
embeddings.embed_query("What's our Q1 revenue?")
```

When editing the same code in Visual Studio Code with the Pylance extension enabled, following error is encountered.
image

In this case, the parameters `task_type`, `google_api_key`, `client_options`, `transport`, and `request_options` are clearly optional, as indicated by the `None` type. `client` does not have the `None` type, but is also optional since it is not provided in the example code snippet.

Note that Pylance does not flag the absence of the `credentials` field, even though it is also optional, and that the only field with a _recognized_ default value (`= None`) is `credentials`.

## Causes

Upon closer examination of the code, it becomes evident that some definitions for Pydantic fields are not handled by Pylance as intended. Examining the definition of [`GoogleGenerativeAIEmbeddings`](https://api.python.langchain.com/en/latest/embeddings/langchain_google_genai.embeddings.GoogleGenerativeAIEmbeddings.html) from the previous example (some methods, field descriptions and doc string for class are redacted for the sake of brevity):

```python
class GoogleGenerativeAIEmbeddings(BaseModel, Embeddings):
    # <doc string>

    client: Any  #: :meta private:
    model: str = Field(
        ...,
        description="...",
    )
    task_type: Optional[str] = Field(
        None,
        description="...",
    )
    google_api_key: Optional[SecretStr] = Field(
        None,
        description="...",
    )
    credentials: Any = Field(
        default=None,  # <- notice the default keyword
        exclude=True,
        description="...",
    )
    client_options: Optional[Dict] = Field(
        None,
        description="...",
    )
    transport: Optional[str] = Field(
        None,
        description="...",
    )
    request_options: Optional[Dict] = Field(
        None,
        description="...",
    )

    @root_validator()
    def validate_environment(cls, values: Dict) -> Dict:
        """Validates params and passes them to google-generativeai package."""
        google_api_key = get_from_dict_or_env(
            values, "google_api_key", "GOOGLE_API_KEY"
        )
        client_info = get_client_info("GoogleGenerativeAIEmbeddings")

        values["client"] = build_generative_service(
            credentials=values.get("credentials"),
            api_key=google_api_key,
            client_info=client_info,
            client_options=values.get("client_options"),
        )
        return values


    # <other methods>
```

we can see that except for `client` and `model` all other fields are given a default value by [`Field`](https://docs.pydantic.dev/latest/api/fields/#pydantic.fields.Field). However, the only default value recognized by Pylance is the one for `credentials`. And it is no coincidence that `credentials` is also the only field where the default value is defined using the `default` **keyword** argument.

It is safe to say that the errors related to `task_type`, `google_api_key`, `client_options`, `transport`, and `request_options` are caused by the default value for them not being provided with the keyword arguments. (Unfortunately, I am not sure why this is the case.)

For the `client` case, we can assume that it's default value is `None`, since it's set within the `validate_environment` method. This case is not explicitly declared, which makes Pylance angry.

### TL;DR
The problem seems to be caused by two different cases:

1. The default value for an optional field is not provided (such as `client` in [`GoogleGenerativeAIEmbeddings`](https://api.python.langchain.com/en/latest/embeddings/langchain_google_genai.embeddings.GoogleGenerativeAIEmbeddings.html)).
2. The default value for an optional field is provided without the `default` keyword using `Field`.

## Solution

Find all the field definitions that are causing problems and fix them. 

To find the examples for the first case, I searched for field definitions of type `Any` without a default value, similar to `client` in [`GoogleGenerativeAIEmbeddings`](https://api.python.langchain.com/en/latest/embeddings/langchain_google_genai.embeddings.GoogleGenerativeAIEmbeddings.html). For the second case, I searched for optional fields using the `Optional[.*] = Field(` regex and manually checked for definitions without the `default` keyword.

It was trivial to apply the fix after finding the faulty lines.

### DIFF
--- a/libs/community/langchain_google_community/gmail/create_draft.py
+++ b/libs/community/langchain_google_community/gmail/create_draft.py
@@ -24,11 +24,11 @@ class CreateDraftSchema(BaseModel):
         description="The subject of the message.",
     )
     cc: Optional[List[str]] = Field(
-        None,
+        default=None,
         description="The list of CC recipients.",
     )
     bcc: Optional[List[str]] = Field(
-        None,
+        default=None,
         description="The list of BCC recipients.",
     )
 
--- a/libs/community/langchain_google_community/gmail/send_message.py
+++ b/libs/community/langchain_google_community/gmail/send_message.py
@@ -27,11 +27,11 @@ class SendMessageSchema(BaseModel):
         description="The subject of the message.",
     )
     cc: Optional[Union[str, List[str]]] = Field(
-        None,
+        default=None,
         description="The list of CC recipients.",
     )
     bcc: Optional[Union[str, List[str]]] = Field(
-        None,
+        default=None,
         description="The list of BCC recipients.",
     )
 
--- a/libs/community/langchain_google_community/places_api.py
+++ b/libs/community/langchain_google_community/places_api.py
@@ -30,7 +30,7 @@ class GooglePlacesAPIWrapper(BaseModel):
     """
 
     gplaces_api_key: Optional[str] = None
-    google_map_client: Any  #: :meta private:
+    google_map_client: Any = None  #: :meta private:
     top_k_results: Optional[int] = None
 
     class Config:
--- a/libs/community/langchain_google_community/search.py
+++ b/libs/community/langchain_google_community/search.py
@@ -47,7 +47,7 @@ class GoogleSearchAPIWrapper(BaseModel):
 
     """
 
-    search_engine: Any  #: :meta private:
+    search_engine: Any = None  #: :meta private:
     google_api_key: Optional[str] = None
     google_cse_id: Optional[str] = None
     k: int = 10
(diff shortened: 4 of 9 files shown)


lkuligin:
@orkhank thanks for your contribution! could fix the remaining linter issues, please?

lkuligin (approved):

PR #365, opened 2024-07-12
Fix: BigQueryVectorStore crashes on BQ tables with more than 5000 rows

### LINKED ISSUE #364: `BigQueryVectorStore` `initialize_bq_vector_index` accesses non-existing key `__last_index_check`
The pydantic `root_validator` accesses a non-existing key `__last_index_check` here:
https://github.com/langchain-ai/langchain-google/blob/main/libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py#L139

Steps to reproduce:
1. Instantiate a `BigQueryVectorStore` on an existing table with more than 5000 rows to trigger indexing (see https://github.com/langchain-ai/langchain-google/blob/main/libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py#L21)
2. An exception occurs

Consequences: `BigQueryVectorStore` is not usable on tables with more than 5000 rows.

From what i could find, there is another `root_validator` found in the [BaseBigQueryVectorStore](https://github.com/langchain-ai/langchain-google/blob/main/libs/community/langchain_google_community/bq_storage_vectorstores/_base.py#L116) which does initialize some other private `__` fields, but not `__last_index_check`. I would assume the fix would either be check before access in `initialize_bq_vector_index` or initialize the field in the base `root_validator`?

thomasfitze:
Fixes #364

### DIFF
--- a/libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py
+++ b/libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py
@@ -138,7 +138,7 @@ class BigQueryVectorStore(BaseBigQueryVectorStore):
 
         if "_last_index_check" not in values:
             values["_last_index_check"] = datetime.min
-        
+ 
         if datetime.utcnow() - values["_last_index_check"] < INDEX_CHECK_INTERVAL:
             return values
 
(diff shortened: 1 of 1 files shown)


lkuligin:
@thomasfitze thanks for your contribution! integration tests are green now, could you fix the linter, please?

lkuligin (approved):

PR #401, opened 2024-07-25
Interface update: remove BQ dependencies from BaseVectoreStore

ThibaultScalbert:
This change is clearly segmenting Feature stores from BQ stores.
The goal is to have 2 separate stores that can sync together independently.

That means that when instantiating a feature store, there is no BQ store instantiation.
The user can switch from one to another for each of the needed feature but they are not mixing together interdependently.


lkuligin:
@eliasecchig

eliasecchig:
Already discussed with @ThibaultScalbert offline, can be closed

Issue #399, opened 2024-07-24
Multimodal FC does not work due to langchain incorrectly parsing as string


kardiff18:
I'm trying to do multimodal FC via langchain. I was allowlisted & all works well if i do it in the Vertex SDK.

However, when i do it in Langchain, i found that the GenerateContentRequest is malformed such that the audio bytes are sent via the Part[text] instead of Part[inline_data]. I think this is because when doing FC in langchain, you have to set a ChatPromptTemplate, which always returns a prompt as a string. Now, if that's true, that's an issue. This is because when langchain parses the user's request to create the Part protos, it first checks for string input. If it's string (https://github.com/langchain-ai/langchain-google/blob/ae52375f697fcbf38884dcc30457d61c1baeaf24/libs/vertexai/langchain_google_vertexai/chat_models.py#L194), then it just adds the whole input to `text` rather than proceeding to check if it's media and then adding as `inline_data`. 

The problem additionally might start upstream here: https://github.com/langchain-ai/langchain-google/blob/ae52375f697fcbf38884dcc30457d61c1baeaf24/libs/vertexai/langchain_google_vertexai/chat_models.py#L279 . This is because here we are _not_ expecting a nested HumanMessage object; however, with my code below, this is actually the case (although the innermost object is technically a string that just needs to be converted back to a Human Message. )

Here is sample code displaying the problem:

```
prompt_template = ChatPromptTemplate.from_messages(
    [
        ("system", SYSTEM_PROMPT),
        ("human", "{input}"),
        ("placeholder", "{agent_scratchpad}"),
    ]
)
agent = agents.create_tool_calling_agent(
    llm,
    tools,
    prompt_template
)
agent_executor = agents.AgentExecutor(agent=agent, tools=tools, verbose=False)
ai_response = agent_executor.invoke({"input":
  [HumanMessage(content=[
            {
                "type": "media",
                "mime_type": "audio/wav",
                "data": bytes

            }
```

lkuligin:
we typically support only publicly available features

in this case, please, feel free to send a PR, but we'll need to add an integration test and allowlist the project we use for integration testing then

PR #404, opened 2024-07-25
Add Multimodal fc support

kardiff18:
fix: https://github.com/langchain-ai/langchain-google/issues/399


libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -235,6 +244,107 @@ def test_multimodal_media_inline_base64(file_uri, mime_type) -> None:
     assert isinstance(output.content, str)
 
 
+@pytest.mark.release
+@pytest.mark.parametrize("file_uri,mime_type", multimodal_inputs)
+def test_multimodal_media_inline_base64_template(file_uri, mime_type) -> None:
+    llm = ChatVertexAI(model_name="gemini-1.5-pro-preview-0514")
```
lkuligin, line 250: why preview model here? we should start getting away from it

kardiff18: fair! not sure why i did it either. updated.

lkuligin (approved):

PR #405, opened 2024-07-25
vertexai[patch]: bind tools in oai format

baskaryan:
Bind tools in openai format so that they can be better traced (protobufs aren't well handled by most tracers, and gemini 
tool format doesn't match most other models).


efriis (approved):
important! ty for fixing

Issue #409, opened 2024-07-30
ChatGoogleGenerativeAI function calling fails with enums in schema


gramliu:
There's an issue with function calling serialization for Enum fields in Pydantic models. Running the code snippet below fails with the following error:
```
ChatGoogleGenerativeAIError('Invalid argument provided to Gemini: 400 * GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[color].type: must be specified\n')Traceback (most recent call last):
```

Minimal reproducible code snippet:
```py
from langchain_google_genai import ChatGoogleGenerativeAI
from langchain_core.pydantic_v1 import BaseModel, Field
from langchain_core.tools import tool
from enum import Enum
from dotenv import load_dotenv

load_dotenv()

ColorEnum = Enum("Color", ["red", "green", "blue"])

class ColorSchema(BaseModel):
  color: ColorEnum = Field(description="The color of the picture.")

@tool("select_color", args_schema=ColorSchema)
def select_color(color: ColorEnum) -> bool:
  """Select a color."""
  return True

gemini_pro = ChatGoogleGenerativeAI(model="gemini-1.5-pro", temperature=0, max_tokens=1000)
gemini_pro = gemini_pro.bind_tools([select_color], tool_choice="select_color")
gemini_pro.invoke("What's your favorite color'?")
```

PR #410, opened 2024-07-31
Add partial support for enums and type unions

gramliu:
https://github.com/langchain-ai/langchain-google/issues/409

This adds partial support for the `allOf` tag produced by pydantic schemas for enums.


libs/genai/langchain_google_genai/_function_utils.py
```diff
@@ -109,6 +109,17 @@ def _format_json_schema_to_gapic(schema: Dict[str, Any]) -> Dict[str, Any]:
             continue
         elif key in ["type", "_type"]:
             converted_schema["type"] = str(value).upper()
+        elif key == "allOf":
```
lkuligin, line 112: It looks to me this implementation is a little bit cleaner:
https://github.com/langchain-ai/langchain-google/blob/e0c17f07bda807cdad97e9f5523a3093cca6f579/libs/vertexai/langchain_google_vertexai/functions_utils.py#L95

wdyt? Do you mind taking the same approach, please?

PR #420, opened 2024-08-03
genai: update pyproject as vertexai

nobu007:
This PR aligns the pyproject.toml files for genai and vertexai, reducing maintenance costs.

## Changes
- packages is almost based by vertexai(especially types-protobuf helpful).
- version is latest of genai and vertexai.
- fix mypy ignore.
- minar fix about SecretStr of test.


libs/genai/pyproject.toml
```diff
@@ -46,6 +54,10 @@ codespell = "^2.2.0"
 optional = true
 
 [tool.poetry.group.test_integration.dependencies]
+numexpr = { version = "^2.8.8", python = ">=3.9,<4.0" }
+google-api-python-client = "^2.114.0"
+google-cloud-datastore = "^2.19.0"
```
lkuligin, line 59: we don't use this library in this project

libs/genai/pyproject.toml
```diff
@@ -46,6 +54,10 @@ codespell = "^2.2.0"
...
+langchain = { git = "https://github.com/langchain-ai/langchain.git", subdirectory = "libs/langchain" }
```
lkuligin, line 60: same here, why do we need this change?

libs/genai/pyproject.toml
nobu007: @lkuligin 
I updated PR. Remove unused libs.

libs/genai/pyproject.toml
nobu007: @lkuligin 
I updated PR. Remove unused libs.

PR #424, opened 2024-08-07
Update drive.py to only use drive.file scope

walterheck:
the drive.readonly scope gives very broad access to an application and is therefore a restricted scope. Restricted scopes are a pain because it means the application using it needs to go through a google restricted scope security review (see: https://developers.google.com/identity/protocols/oauth2/production-readiness/restricted-scope-verification).

With this change, the scope gets reduced to a recommended scope (see: https://developers.google.com/drive/api/guides/api-specific-auth)

### DIFF
--- a/libs/community/langchain_google_community/drive.py
+++ b/libs/community/langchain_google_community/drive.py
@@ -15,7 +15,7 @@ from langchain_core.document_loaders import BaseLoader
 from langchain_core.documents import Document
 from langchain_core.pydantic_v1 import BaseModel, root_validator, validator
 
-SCOPES = ["https://www.googleapis.com/auth/drive.readonly"]
+SCOPES = ["https://www.googleapis.com/auth/drive.file"]
 
 
 class GoogleDriveLoader(BaseLoader, BaseModel):


lkuligin (approved):

oscarfesa:
After some debugging in my code (that it used to work) I believe I've identified a potential root cause related to this recent code change.
Specifically, a commit was made that modified the SCOPES variable. Previously, it was set to:

``` SCOPES = ["https://www.googleapis.com/auth/drive.readonly"]```
This scope allowed read-only access to Drive. However, the recent change updated it to:

``` SCOPES = ["https://www.googleapis.com/auth/drive.file"]```

This scope cannot load content from Google Documents when using in combination with ```service_account_key```. It throws the following exception:

 `<HttpError 404 when requesting https://www.googleapis.com/drive/v3/files/XXXXXXXXX?supportsAllDrives=true&fields=modifiedTime%2Cname&alt=json returned "File not found: XXXXXXXXX.". Details: "[{'message': 'File not found: XXXXXXXXX.', 'domain': 'global', 'reason': 'notFound', 'location': 'fileId', 'locationType': 'parameter'}]">`
 
SCOPES variable is hardcoded in drive.py so I cannot override the value. I would suggest to revert the change or adding SCOPES as a parameter to GoogleDriverLoader constructor.

**Comment by walterheck** [comment]
Reverting this change doesn't seem to be the way, it prevents anyone from making an app public, the drive.readonly scope is a restricted scope that google doesn't recommend using and requires a weeks-long in-depth review before you can take an app public. 

I do agree that there are usecases for both so it would be best to make this scope overrideable according to the usecase.

**Comment by oscarfesa** [comment]
Okay, I understand the reasoning behind using the recommended scope. But there might be some considerations to keep in mind.

For example, this code can no longer load the document content because of the change in 'scope'. So, the Loader is basically useless for my purposes.
```
loader = GoogleDriveLoader(
                    service_account_key='path/to/my/sa_key.json',
                    document_ids=['XXXXXXX']
                )
document = loader.load()
```
I'm forced to either stay on version 1.0.7 or create a completely new loader from scratch (as I cannot modify the scope to make it work)

I understand that app verification processes are used when creating a public app that acts on behalf a user. However, in cases where a service account is used the approach is different. My use case is an internal pipeline for processing company's corporate information where there isn't even a need to publish an app. It is just a pipeline with server-to-server communications. 

I don't know if there's something I'm missing... At least, when using service accounts, I think the scope could be the previous one by default. Currently I am not able to see any way to make this loader work...

Raj725:
I agree with @oscarfesa. 
Adding SCOPES as a parameter to the GoogleDriveLoader constructor is much needed. 

Existing apps have stopped working after upgrading `langchain-google-community`, and the worst part is that it doesn’t throw any errors; it just quietly returns an empty list of documents. We were using it for an internal app.

walterheck:
adding the scope is needed indeed, to make both use cases work. I have actually already moved to use the google drive loader made by @pprados: https://github.com/pprados/langchain-googledrive
That one allows you to pass in a credentials object which already has the scope needed, removing the whole need for a library to make a decision on this.

Raj725:
@walterheck Thank you for sharing this advanced version! I'll definitely give it a try.

oscarfesa:
Just as a comment. Shortly after writing my previous comment I also found the same library and started to use it. So far so good. 
There is a lot of additional functionality and it leaves this library far behind.

PR #428, opened 2024-08-07
fix: Standardize default parameters across Gemini and Gemini Advanced models

mitraan-deshaw:
Currently, models in the `GEMINI` family have no default parameters set, whereas they are set for all other Google models. This implies models in the new `GEMINI_ADVANCED` family also have default parameters set, such `max_output_tokens=128` (which are meant for PaLM and Bison models). 

This PR aims to fix that by making the default parameter behavior of `GEMINI` and `GEMINI_ADVANCED` model families identical for now. We can update defaults for each model family in the future.


mitraan-deshaw:
I see a couple of integration tests failing, but they don't seem to be related to the changes I made. I'm happy to update the tests if needed.

lkuligin (approved):

PR #450, opened 2024-08-19
vertex: fix SafetySetting import

kardiff18:
## PR Description

Fixes the import of SafetySetting, adds to existing unit test for import checks

## Relevant issues

N/A

## Type

🐛 Bug Fix
✅ Test


lkuligin (approved):

libs/vertexai/tests/integration_tests/test_llms.py
```diff
@@ -10,7 +10,7 @@
 from langchain_google_vertexai.llms import VertexAI
 from tests.integration_tests.conftest import _DEFAULT_MODEL_NAME
 
-model_names_to_test = ["text-bison@001", _DEFAULT_MODEL_NAME]
+model_names_to_test = ["gemini-1.0-pro", _DEFAULT_MODEL_NAME]
```
lkuligin, line 13: it's enough to just use _DEFAULT_MODEL_NAME, isn't it?

kardiff18: I found that the integration for `test_stream` was failing in my env... since I didn't make any changes that would cause this (IMO), I looked into it. I believe test-bison@001 was discontinued already, so I figured I should switch it to Gemini. When I did this, the test succeeded. 

Happy to change my PR to get rid of the first element in the array entirely if you wish, I just figured I should update the existing text-bison model since it's so old. LMK what you think @lkuligin

PR #466, opened 2024-08-25
vertexai: Show tqdm progress bar for VertexAIEmbeddings

huzaifa50:
## PR Description
In the [deprecated version of Vertex AI Embedding](https://api.python.langchain.com/en/latest/_modules/langchain_community/embeddings/vertexai.html#VertexAIEmbeddings) option to show progress of batches was present. This PR add the same option in the newer version

## Type

🆕 New Feature


lkuligin:
I don't like the idea of bringing in tqdm, maybe we could just use logger to inform about the progress instead in stdout? @baskaryan @efriis do you folks add it for other LangChain integrations?

efriis:
we do not

PR #472, opened 2024-09-03
genai, vertex, community[patch]: backward-compatible pydantic v2 changes

ccurme:


ccurme:
Vertex integration tests passed and then started to fail following an innocuous-looking change to langchain-google-community. I don't have access to GCloud logs for the job. Vertex integration tests pass locally. Going ahead with merge.

PR #473, opened 2024-09-05
langchain-google-vertexai: Update vision_models.py to allow kwargs usage

TommasoPetrolito:
## Checklist for PR Creation

- [x] PR Title: "langchain-google-vertexai: Update vision_models.py to allow kwargs usage"

- [x] PR Description and Relevant issues:

  - This change touches `_prepare_params()` methods in order to avoid that keys not included in default parameters are filtered out systematically
```
def _prepare_params(self, **kwargs: Any) -> Dict[str, Any]:
        params = self._default_params
        for key, value in kwargs.items():
             if key in params and value is not None:
                    params[key] = value
```
changed to:
```
def _prepare_params(self, **kwargs: Any) -> Dict[str, Any]:
        params = self._default_params
        for key, value in kwargs.items():
             if value is not None:
                   params[key] = value
```

- This change also touches all the final calls to vertexai image related functions when they take into account the additional parameters, they were only taking into account, for some reason, `messages[0].additional_kwargs`, instead `kwargs` coming from outer functions calls were not taken into account and have been added:
```
        image_str_list = self._generate_images(
            prompt=user_query, **messages[0].additional_kwargs
        )
```
changed to:
```
        image_str_list = self._generate_images(
            prompt=user_query, **messages[0].additional_kwargs, **kwargs
        )
```

  - Relevant issues: It was not possible to use kwargs like `{"safety_filter_level": "block_few", "person_generation": "allow_all"}` from `langchain-core` `BaseChatModel` `invoke()` method. They got filtered out and never reached the final image generation call.
  
  - No specific dependencies required for this fix.

- [ ] Add Tests and Docs:

  - None

- [x] Lint and Test:
  - The amount of changes is limited to very specific points and should not require lint check, as you can see
  - I have tested it by directly pip installing from this commit hash

## PR Description

This PR addresses the impossibility to use additional kwargs with models classes' methods inside langchain_google_vertexai/vision_models.py. E.g.: It was not possible to pass parameters like  `{"safety_filter_level": "block_few", "person_generation": "allow_all"}` to the image generation, these parameters used to get filtered out systematically.

## Type

🐛 Bug Fix


TommasoPetrolito:
Alternatively, instead of disabling the filtering out of non-default keys contained inside self._default_params, all the useful parameters should be added to default_params accordingly.
Of course this alternative approach would require continuous timely update of the list of default params.
Both options are OK and would allow library users to effectively use kwargs.

lkuligin:
Thanks for your contribution! 
I'm not fully sure it's an ideal way of doing things. An alternative would be to add additional keys to default_params with None values (to be tested, but typically protos on the server side handle None values well). Since after this change if a user includes into `kwargs` an argument that is not supported by the proto, the model will fail (since this argument will be always passed further).

@efriis @baskaryan  What is the general approach, should we be graceful to any argument in **kwargs that is not accepted by the underlying model, or should we fail? Maybe I'm overthinking here.

TommasoPetrolito:
> Thanks for your contribution! 
> I'm not fully sure it's an ideal way of doing things. An alternative would be to add additional keys to default_params with None values (to be tested, but typically protos on the server side handle None values well). Since after this change if a user includes into `kwargs` an argument that is not supported by the proto, the model will fail (since this argument will be always passed further).
> 
> @efriis @baskaryan  What is the general approach, should we be graceful to any argument in **kwargs that is not accepted by the underlying model, or should we fail? Maybe I'm overthinking here.

I totally agree, I am not sure as well.
It depends on what is the exact purpose of default_params, I might have interpreted it wrong.
If it is there to automatically fill them with default values that are only eventually replaced by kwargs if a key is there, then filtering out additional keys might be an undesired behavior, just a leftover.
If default_params is more like a supported_params, then yes, it makes more sense to add all the supported keys to default_params.
I think any solution allowing to use at least the actually supported params (adding them is perfect) would work fine from a library-user POV, anything is better then just having the additional params just silently ignored, I think (even a crash or error would be better so that the user is at least aware of whether or not them are taken into account). From the mantaining POV, adding all supported params to default_params means possibly editing this code every time vertexai library and model handles new parameters or drops a parameter, while accepting all of them without filtering them out, only puts the responsibility of passing actually existing parameters on the user side.

lkuligin:
@TommasoPetrolito feel free to take a look how it works here, please:
https://github.com/langchain-ai/langchain-google/blob/895822116bb92fa256c1da36d700754e2f7782b5/libs/vertexai/langchain_google_vertexai/chat_models.py#L605

maybe we could align the approach, wdyt?

TommasoPetrolito:
> @TommasoPetrolito feel free to take a look how it works here, please:
> 
> https://github.com/langchain-ai/langchain-google/blob/895822116bb92fa256c1da36d700754e2f7782b5/libs/vertexai/langchain_google_vertexai/chat_models.py#L605
> 
> maybe we could align the approach, wdyt?

Hello, thank you again for your feedback, it is not 100% clear to me what you mean here. Do you suggest to proceed on a code change approach involving checks on the supported parameters that imitates / is coherently aligned with the code as done here: 

> https://github.com/langchain-ai/langchain-google/blob/895822116bb92fa256c1da36d700754e2f7782b5/libs/vertexai/langchain_google_vertexai/chat_models.py#L605
>  
?

Or you mean that we should discuss the approach a bit further here and align our views on the approach after a discussion?

I can, for sure, change my code proposal inside vision_models.py to include a supported_params specific variable/property for each vision model (listing all supported params already handled by vertexai library and models) and implement a check with filtering based on those variable/property instead of default_params, that would then be used only for the automatic filling-in of default values for those default_params key=value pairs, not to filter out params.

What do you think?

Thank you again.

lkuligin (approved):

Issue #392, opened 2024-07-23
index out out range exception ChatVertexAI - fix suggested in the ticket


gajendrarao1979:
Hi,

We are getting following error while using ChatVertexAI (langchain-google-vertexai 1.0.7) 

  ```
File "c:\Users\gajen\anaconda3\envs\.venvRasa3Actions\lib\site-packages\langchain_core\language_models\chat_models.py", line 412, in astream
    async for chunk in self._astream(
  File "c:\Users\gajen\anaconda3\envs\.venvRasa3Actions\lib\site-packages\langchain_google_vertexai\chat_models.py", line 1535, in _astream
    request = self._prepare_request_gemini(messages=messages, stop=stop, **kwargs)
  File "c:\Users\gajen\anaconda3\envs\.venvRasa3Actions\lib\site-packages\langchain_google_vertexai\chat_models.py", line 1202, in _prepare_request_gemini
    system_instruction, contents = _parse_chat_history_gemini(messages)
  File "c:\Users\gajen\anaconda3\envs\.venvRasa3Actions\lib\site-packages\langchain_google_vertexai\chat_models.py", line 300, in _parse_chat_history_gemini
    prev_content = vertex_messages[-1]
IndexError: list index out of range

```

This is happening because currently library assumes that first message will be either Human Message or System Message which is not the case for our use case (we put System Message as suffix message for certain scenario).

![image](https://github.com/user-attachments/assets/2b8d605e-e284-404f-abea-4711d21c26f3)

The fix is simple all the lines from 300 till 307 should be moved under if statement (if len(vertex_messages) > 0:). as follows 
![image](https://github.com/user-attachments/assets/4f2ba99f-9535-43bb-86f3-bc2f15be63b2)

tnguyen2907:
Hi,

Has this problem been fixed yet? I have updated to langchain-google-vertexai 1.0.10 but the same error still happened. For me, only HumanMessage at the beginning works.

`  File "C:\Users\trung\anaconda3\envs\obot\Lib\site-packages\langchain_google_vertexai\chat_models.py", line 1165, in _generate
    return self._generate_gemini(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\trung\anaconda3\envs\obot\Lib\site-packages\langchain_google_vertexai\chat_models.py", line 1321, in _generate_gemini
    request = self._prepare_request_gemini(messages=messages, stop=stop, **kwargs)
[...]

PR #486, opened 2024-09-11
fix: index error when first message is AIMessage in _parse_chat_history

sataycat:
## PR Description

Fixes https://github.com/langchain-ai/langchain-google/issues/392

## Type

🐛 Bug Fix


lkuligin:
thanks for your contribution! I believe it was fixed here: https://github.com/langchain-ai/langchain-google/pull/496 (in a slightly different manner).

Do you think it solved the problem?

PR #495, opened 2024-09-16
vertexai: Support list URI input to `embed_image()` API

duwenxin99:
- Other embedding services are supporting list of URI input to the image embedding API `embed_image()`. Replacing the current `embed_image()` API with `embed_images()` that takes a list of image URIs as input allows it to be used with the [LangChain vector store and similarity search integrations](https://github.com/googleapis/langchain-google-alloydb-pg-python/pull/207) that are currently being built for Google Cloud databases.


libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -470,27 +470,31 @@ def embed_query(self, text: str) -> List[float]:
 
     def embed_image(
         self,
-        image_path: str,
+        uris: List[str],
```
lkuligin, line 473: it seems like a breaking change to me, can we add a deprecation warning instead and implement a new method maybe?

libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -470,27 +470,31 @@ def embed_query(self, text: str) -> List[float]:
...
         image_loader = ImageBytesLoader()
-        bytes_image = image_loader.load_bytes(image_path)
-        image = Image(bytes_image)
-        result: MultiModalEmbeddingResponse = self.instance[
-            "get_embeddings_with_retry"
-        ](image=image, contextual_text=contextual_text, dimension=dimensions)
-        return result.image_embedding
+        embeddings = []
+        for image_path in uris:
+            bytes_image = image_loader.load_bytes(image_path)
```
lkuligin, line 493: do embeddings models support pointers to GCS actually?

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/embeddings.py
duwenxin99: Currently other embedding models mostly only support local URIs.

libs/vertexai/langchain_google_vertexai/embeddings.py
duwenxin99: added deprecation warning and new method

libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -498,5 +504,33 @@ def embed_image(
         ](image=image, contextual_text=contextual_text, dimension=dimensions)
         return result.image_embedding
 
+    def embed_images(
+        self,
+        uris: List[str],
+        contextual_text: Optional[str] = None,
+        dimensions: Optional[int] = None,
+    ) -> List[float]:
```
lkuligin, line 512: should it be List[List[float]] maybe?

duwenxin99: Thanks for catching that. Updated.

lkuligin (approved):

libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -487,6 +488,12 @@ def embed_image(
         Returns:
             Embedding for the image.
         """
+        warnings.warn(
```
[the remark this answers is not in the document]
duwenxin99: Added. Is there a specific removal version you'd like?

lkuligin (approved):

libs/vertexai/langchain_google_vertexai/embeddings.py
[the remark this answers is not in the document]
lkuligin: all good, thx!

Issue #492, opened 2024-09-15
Gemini ai error " 400 * GenerateContentRequest.tools[0].function_declarations[3].parameters.properties[actual_departure].type: must be specified "


AtmehEsraa:
I tried to run the Customer Support Bot notebook, and it gave me an error message. Could anyone help me?
[Langchain_Evaluate_Gemini_SA_on_Flight_Booking,_RAG_and_Chat_Task_All_Together (1).zip](https://github.com/user-attachments/files/17005226/Langchain_Evaluate_Gemini_SA_on_Flight_Booking._RAG_and_Chat_Task_All_Together.1.zip)

PR #508, opened 2024-09-22
genai: fix _function_utils.py for shell tools

nobu007:
## PR Description
genai: fix _function_utils.py for shell tools

## Relevant issues

Not sure but may someting related with.

https://github.com/langchain-ai/langchain-google/issues/492

## Type

🐛 Bug Fix


lkuligin:
should we close it maybe?

PR #515, opened 2024-09-25
vertexai[patch]: exclude full_model_name from ser

baskaryan:
there's already project, location and model. no need for full_model_name to be serialized


lkuligin:
please, ignore `test_maas.py` failing (I'd mark them as `fail` if I won't find the issue quickly, or feel free to do so yourself :))

lkuligin (approved):

Issue #503, opened 2024-09-20
[Feature Request] ChatVertexAI: Support for JSON Schema in .withStructuredOutput Method


luonghuuthanhnam:
Currently, the `.withStructuredOutput` method in `ChatVertexAI` only supports the `functionCalling` mechanism. However, Google VertexAI already allows for the use of JSON Schema through model configuration. Are there any plans to support `jsonSchema` in this method?

lkuligin:
You can already use controlled generation with `response_mime_type` parameter.
Or how would you see this?

luonghuuthanhnam:
> You can already use controlled generation with `response_mime_type` parameter. Or how would you see this?
Thank for your reply.
I mean the `response schema`.
`response_mime_type` specifies the format of the response (e.g., JSON, text), while the `response schema` defines the structure and data types expected in the response.
[Ref](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output)

Issue #345, opened 2024-07-02
Specifying json_mode in with_structured_output for Custom Schemas in LangChain


Ben81828:
Is it possible to pass in our defined schema and specify the use of json_mode, like in this example?
https://python.langchain.com/v0.1/docs/modules/model_io/chat/structured_output/

```python
structured_llm = model.with_structured_output(Joke, method="json_mode")
```

lkuligin:
It sounds to me like a good improvement.
Would you be open to send a PR for that?

haveamission:
@lkuligin Do you still need that PR? Seems like it was merged into the library?

alexminza:
This should also be added to `ChatGoogleGenerativeAI`

Reference: https://ai.google.dev/gemini-api/docs/structured-output

mdrxy:
@alexminza feel free to open a new issue if it is not already resolved

PR #469, opened 2024-08-28
genai: fix pydantic structured_output with array

nobu007:
## PR Description

Fix pydantic structured_output with array

## Relevant issues

#24225
https://github.com/langchain-ai/langchain/issues/24225

## Type

🐛 Bug Fix

## Testing(optional)

This PR can pass like here.

```
class TestClass(BaseModel):
    """_summary_

    Args:
        BaseModel (_type_): _description_
    """
    test_val: str = Field(default="", description="value for test(works).")
    test_val_list: List[str] = Field(
        default_factory=list,
        description="values for test(not works).",
    )
```

Test result
```
jinno@jinno-desktop:~/git/drill/gamebook/codeinterpreter_api_agent/langchain-google/libs/genai$ poetry run pytest --extended tests/integration_tests/test_chat_models.py::test_chat_vertexai_gemini_function_calling
============================================================== test session starts ===============================================================
platform linux -- Python 3.10.10, pytest-7.4.4, pluggy-1.5.0
rootdir: /home/jinno/git/drill/gamebook/codeinterpreter_api_agent/langchain-google/libs/genai
configfile: pyproject.toml
plugins: anyio-4.4.0
collected 1 item                                                                                                                                 

tests/integration_tests/test_chat_models.py .                                                                                              [100%]

================================================================ warnings summary ================================================================
../../../../../../../.cache/pypoetry/virtualenvs/langchain-google-genai-ov3b6nnP-py3.10/lib/python3.10/site-packages/_pytest/config/__init__.py:1373
  /home/jinno/.cache/pypoetry/virtualenvs/langchain-google-genai-ov3b6nnP-py3.10/lib/python3.10/site-packages/_pytest/config/__init__.py:1373: PytestConfigWarning: Unknown config option: asyncio_mode
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```
## Note1

This code shoud be updated later.
_set_schema_items() can't create correct function-calling scema.
https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/function-calling?hl#schema

## Note2

This PR may be fix https://github.com/langchain-ai/langchain-google/issues/492.
integration_test filed by this issue.

PR #533, opened 2024-10-04
vertexai: Allow json_mode in with_structured_output

### LINKED ISSUE #531: [Feature Request] Use proper JSON Mode for Gemini Models instead of Tool calling
It seems `with_structured_output` in current `ChatVertexAI` is implemented using tool calling (`JsonOutputToolsParser`) instead of JSON mode. Note that these are different things. Gemini now supports JSON mode as documented [here](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/control-generated-output).

jzaldi:
## PR Description

Adds a keyword parameter `method='json_mode'` to `with_structured_output` in `ChatVertexAI` to use controlled generation rather than function calling.

## Relevant issues

Fixes #531, #503, #345,  #144

## Type

🆕 New Feature

## Example

```
from pydantic import BaseModel

class Response(BaseModel):
    response: str
    reason: str

model = (
    ChatVertexAI(model_name="gemini-1.5-pro-001")
    .with_structured_output(Response, method="json_mode")
)

model.invoke("What is the color of the sea?")

>> Response(response="The sea doesn't have a single color. It often appears blue due to the reflection of the sky, but it can also be green, turquoise, or even brown depending on depth, sediment, and the presence of microorganisms.", reason='The color of the sea is determined by the way it absorbs and reflects light, not an inherent pigment.')

```


jzaldi:
We have an issue with nested pydantic models. The schema generated contains relative references and gemini api doesn't support that. I'll implement a postprocessing function to get rid of it but is not ideal.

jzaldi:
Also I don't see any conflicts with what is being implemented in  #469

jzaldi:
Now we are only missing self referencing objects, something like this doesn't work and I don't think we can do anything about it if gemini's API doesn't support references:

```
class Status(BaseModel):
    substatus: "Status"
```

alexminza:
This should also be added to `ChatGoogleGenerativeAI`

Reference: https://ai.google.dev/gemini-api/docs/structured-output

PR #534, opened 2024-10-04
vertexai: Include avg_logprobs in response metadata

### LINKED ISSUE #525: Add support for AvgLogprobs to VertexAI integrations
Hi LangChain Team!

Can we please add support for returning AvgLogprobs from calls to VertexAI models please?

Thank you!

Regards
Jasen

jzaldi:
## PR Description

Includes `avg_logprobs` in the response metadata

## Relevant issues

Fixes #525

## Type

🆕 New Feature

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_utils.py
@@ -176,8 +176,11 @@ def get_generation_info(
             "finish_reason": (
                 candidate.finish_reason.name if candidate.finish_reason else None
             ),
-            "avg_logprobs": candidate.avg_logprobs,
         }
+
+        if hasattr(candidate, "avg_logprobs") and candidate.avg_logprobs is not None:
+            info["avg_logprobs"] = candidate.avg_logprobs
+
         try:
             if candidate.grounding_metadata:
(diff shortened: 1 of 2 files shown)


libs/vertexai/langchain_google_vertexai/_utils.py
```diff
@@ -176,6 +176,7 @@ def get_generation_info(
             "finish_reason": (
                 candidate.finish_reason.name if candidate.finish_reason else None
             ),
+            "avg_logprobs": candidate.avg_logprobs,
```
lkuligin, line 179: is it exposed for all models now? should we check whether the attribute exists maybe?

jzaldi: It's just for Gemini, is an attribute of the class of the candidates so idk we can check if the attribute exists or maybe doesn't include it if it's None

lkuligin: yep, like at the line above, plaese

jzaldi: Added the checks

lkuligin (approved):

jaymon0703:
Thank you @lkuligin and @jzaldi

PR #550, opened 2024-10-11
vertexai: fix avg log probs nan values

eliasecchig:
Following up on https://github.com/langchain-ai/langchain-google/pull/534/files

Fixes behaviour when using astream_events and stream methods, which requires merging together metadata where some of the chunks might contain "nan" Float values.

e.g
```
left = {'avg_logprobs': 0.0, 'safety_ratings': []}
others = {'avg_logprobs': nan, 'safety_ratings': []}
```

E                   TypeError: Additional kwargs key avg_logprobs already exists in left dict and value has unsupported type <class 'float'>.

We introduce a generic integration test with langraph to catch this kind of errors in the future for any metadata field.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_utils.py
@@ -1,6 +1,7 @@
 """Utilities to init Vertex AI."""
 
 import dataclasses
+import math
 import re
 from enum import Enum, auto
 from importlib import metadata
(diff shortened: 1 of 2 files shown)


libs/vertexai/langchain_google_vertexai/_utils.py
```diff
@@ -179,7 +180,10 @@ def get_generation_info(
         }
 
         if hasattr(candidate, "avg_logprobs") and candidate.avg_logprobs is not None:
-            info["avg_logprobs"] = candidate.avg_logprobs
+            if isinstance(candidate.avg_logprobs, float) and not math.isnan(
```
lkuligin, line 183: we should probably check if it's >0 too

lkuligin (approved):

PR #558, opened 2024-10-17
[vertexai] Fix Datastore error on indexing large fields

lspataroG:
As the `self._text_property_name` and other fields might contain large text, this fix prevents from indexing such fields that currently triggers an error (see Datastores's maximum size of an indexed string property [limitation](https://cloud.google.com/datastore/docs/concepts/limits)). This also saves datastore's costs, preventing a non necessary field to be indexed. We can now pass a list `exclude_from_indexes` to the class, which is the same list that the sdk expects in input when creating an entity.

🐛 Bug Fix

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vectorstores/document_storage.py
+++ b/libs/vertexai/langchain_google_vertexai/vectorstores/document_storage.py
@@ -289,7 +289,10 @@ class DataStoreDocumentStorage(DocumentStorage):
 
             entities = []
             for key, document in zip(keys, documents):
-                entity = self._client.entity(key=key, exclude_from_indexes=[self._text_property_name])
+                entity = self._client.entity(
+                    key=key,
+                    exclude_from_indexes=[self._text_property_name]
+                )
                 entity[self._text_property_name] = document.page_content
                 entity[self._metadata_property_name] = document.metadata
                 entities.append(entity)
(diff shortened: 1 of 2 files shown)


lkuligin (approved):

lspataroG:
@lkuligin I generalised the fix by passing the exclusion list in init and fixed some lint issues.

libs/vertexai/langchain_google_vertexai/vectorstores/document_storage.py
```diff
@@ -234,6 +234,7 @@ def __init__(
         kind: str = "document_id",
         text_property_name: str = "text",
         metadata_property_name: str = "metadata",
+        exclude_from_indexes: List[str] = [],
```
lkuligin, line 237: please, try to avoid default mutable values. Let's change it to Optional[List[str]], please, and check for None.

libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
```diff
@@ -378,6 +378,7 @@ def from_components(
         datastore_kind: str = "document_id",
         datastore_text_property_name: str = "text",
         datastore_metadata_property_name: str = "metadata",
+        exclude_from_indexes: List[str] = [],
```
lkuligin, line 381: same as above

PR #559, opened 2024-10-21
vectorstores: add vertex vector search datapoints deletion

Pablito2020:
This PR is very similar to #331, but fully implemented and with metadata filtering support.

Now, you can delete documents from VectorVertexAiSearch with:

## Delete datapoints by id
```python
ids = [] # in case you know the datapoints id's you can add them here!
vector_store.delete(ids)
```

## Delete datapoints via metadata values.

The more powerful feature is the ability to delete datapoints based on metadata. For example, to delete all datapoints originating from a specific source file:
```python
file_to_delete = "testing.pdf"
# This will delete all the datapoints that have indexes from this source file!
vector_store.delete(metadata={"source": file_to_delete})
```

You can filter by any metadata key-value pair, not just the source file. For instance, to delete all the documents 15th page:
```python
# Delete all the 15th pages
vector_store.delete(metadata={"page": 15}) 
```

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
+++ b/libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
@@ -20,6 +20,8 @@ from langchain_google_vertexai.vectorstores._utils import (
     to_data_points,
 )
 
+MAX_DATA_POINTS = 10000
+
 
 class Searcher(ABC):
     """Abstract implementation of a similarity searcher."""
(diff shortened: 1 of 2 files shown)


lkuligin:
@Pablito2020 thanks a lot for your contribution! Could you finalize it and fix lint errors, please?

Pablito2020:
> @Pablito2020 thanks a lot for your contribution! Could you finalize it and fix lint errors, 

The feature is finalized and works as expected, I've just fixed the linter errors.

The CI is complaining about some LLM tests (which I didn't touch) and they don't test the vector search vector store implementation.

lkuligin (approved):

PR #567, opened 2024-10-24
community: add custom embedding functionality to VertexAISearchRetriever

ThibaultScalbert:
## PR Description

Add the option to bring your own embedding to the langchain retriever.
To use it you need to bring the model, the column of the datastore with the embedding and the ratio for the ranking_expression.

## Type

🆕 New Feature
✅ Test

### DIFF
--- a/libs/community/tests/unit_tests/test_vertex_ai_search.py
+++ b/libs/community/tests/unit_tests/test_vertex_ai_search.py
@@ -320,61 +320,98 @@ def test_custom_embedding_with_valid_values() -> None:
     """
     Test with a valid custom embedding model and field path.
     """
-    embeddings = FakeEmbeddings(size=100)
-    retriever = VertexAISearchRetriever(
-        project_id="mock-project",
-        data_store_id="mock-data-store",
-        custom_embedding=embeddings,
-        custom_embedding_field_path="embedding_field",
-        custom_embedding_ratio=0.5,
-    )
-    search_request = retriever._create_search_request(query="query_value")
-    assert search_request.embedding_spec is not None
-    assert search_request.ranking_expression == (
-        "0.5 * dotProduct(embedding_field) + 0.5 * relevance_score"
-    )
+    # Mock the SearchServiceClient to avoid real network calls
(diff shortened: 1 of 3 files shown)


libs/community/tests/unit_tests/test_vertex_ai_search.py
```diff
@@ -313,3 +314,67 @@ def test_convert_unstructured_search_response_extractive_answers(
...
         assert "previous_segments" not in documents[1].metadata
         assert "next_segments" not in documents[1].metadata
+
+
+def test_custom_embedding_with_valid_values() -> None:
+    """
+    Test with a valid custom embedding model and field path.
+    """
+    embeddings = FakeEmbeddings(size=100)
+    retriever = VertexAISearchRetriever(
```
lkuligin, line 324: please, take a look at the example above. You need to mock the SearchServiceClient, otherwise unit tests fail because of no auth on the github worker.

ThibaultScalbert: Thanks, ok, I tried again. I hope it works.

libs/community/tests/unit_tests/test_vertex_ai_search.py
```diff
@@ -5,6 +5,7 @@
 from google.auth import credentials as ga_credentials
 from google.cloud.discoveryengine_v1beta import Document as DiscoveryEngineDocument
 from google.cloud.discoveryengine_v1beta.types import SearchRequest, SearchResponse
+from langchain_community.embeddings import FakeEmbeddings
```
lkuligin, line 8: we try not to depend on langchain_community.
please, import `from langchain_core.embeddings import FakeEmbeddings` instead

lkuligin (changes requested):

libs/community/tests/unit_tests/test_vertex_ai_search.py
ThibaultScalbert: Make sense. Done!

lkuligin (approved):

Issue #355, opened 2024-07-05
Model 'text-unicorn' returns error "ResourceExhausted: 429 Quota exceeded" with langchain-google-vertexai versions >= 1.0.4


maximilienroberti:
While running the following script:
```python
from langchain_google_vertexai import VertexAI
llm = VertexAI(model_name="text-unicorn@001")
llm.invoke("Hello, what is your name?")
```
It prints the following:
```text
Retrying langchain_google_vertexai.llms._completion_with_retry.<locals>._completion_with_retry_inner in 4.0 seconds as it raised ResourceExhausted: 429 Quota exceeded for aiplatform.googleapis.com/generate_content_requests_per_minute_per_project_per_base_model with base model: text-unicorn. Please submit a quota increase request. https://cloud.google.com/vertex-ai/docs/generative-ai/quotas-genai..
Retrying langchain_google_vertexai.llms._completion_with_retry.<locals>._completion_with_retry_inner in 4.0 seconds as it raised ResourceExhausted: 429 Quota exceeded for aiplatform.googleapis.com/generate_content_requests_per_minute_per_project_per_base_model with base model: text-unicorn. Please submit a quota increase request. https://cloud.google.com/vertex-ai/docs/generative-ai/quotas-genai..
Retrying langchain_google_vertexai.llms._completion_with_retry.<locals>._completion_with_retry_inner in 4.0 seconds as it raised ResourceExhausted: 429 Quota exceeded for aiplatform.googleapis.com/generate_content_requests_per_minute_per_project_per_base_model with base model: text-unicorn. Please submit a quota increase request. https://cloud.google.com/vertex-ai/docs/generative-ai/quotas-genai..
Retrying langchain_google_vertexai.llms._completion_with_retry.<locals>._completion_with_retry_inner in 8.0 seconds as it raised ResourceExhausted: 429 Quota exceeded for aiplatform.googleapis.com/generate_content_requests_per_minute_per_project_per_base_model with base model: text-unicorn. Please submit a quota increase request. https://cloud.google.com/vertex-ai/docs/generative-ai/quotas-genai..
[...]

PR #580, opened 2024-10-29
added text-unicon to PALM family

lkuligin:
#355

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_utils.py
@@ -138,7 +138,7 @@ class GoogleModelFamily(str, Enum):
             return GoogleModelFamily.CODEY
         if "medlm-medium@latest" in model_name:
             return GoogleModelFamily.GEMINI
-        if "bison" in model_name or "medlm" in model_name:
+        if "bison" in model_name or "medlm" in model_name or "unicorn" in model_name:
             return GoogleModelFamily.PALM
         return GoogleModelFamily.GEMINI
 


libs/vertexai/langchain_google_vertexai/_utils.py
```diff
@@ -138,7 +138,7 @@ def _missing_(cls, value: Any) -> "GoogleModelFamily":
             return GoogleModelFamily.CODEY
         if "medlm-medium@latest" in model_name:
             return GoogleModelFamily.GEMINI
-        if "bison" in model_name or "medlm" in model_name:
+        if "bison" in model_name or "medlm" in model_name or "unicon" in model_name:
```
gawbul, line 141: Should be `unicorn` here? Could you update, please 🙏

gawbul (changes requested):

libs/vertexai/langchain_google_vertexai/_utils.py
gawbul: Tried to update myself but I don't have the permissions to push.

gawbul: @lkuligin FYI 😄

lkuligin: sorry, fixed

gawbul:
Looks like some tests are failing, @lkuligin?
Not sure if they just need re-running, or if anything is broken with the change?
I unfortunately can't see the full build log on Google Cloud.

PR #595, opened 2024-11-08
vertexai: add finish message when available to the response object from gemini …

nanodan:
## PR Description

Currently the response from gemini is parsed into a dict which is passed as the `generation_info` to a `ChatGeneration` object. However, this is missing a key piece of information useful for debugging when function calls fail when using function calling via the `bind_tools` syntax which is the `finish_message`. The finish message is populated with information about the malformed function call in the case of `MALFORMED_FUNCTION_CALL` errors. This PR simply adds this field when available to the dictionary passed back.

## Relevant issues

None

## Type

🆕 New Feature

## Changes(optional)

Added in `finish_message` field to the `generation_info` in the case of a gemini model invocation.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_utils.py
@@ -177,6 +177,9 @@ def get_generation_info(
             "finish_reason": (
                 candidate.finish_reason.name if candidate.finish_reason else None
             ),
+            "finish_message": (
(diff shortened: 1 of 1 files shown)


nanodan:
I can't see the results of the build to see why this is failing, all my local tests complete successfully, and this is a very minor change.

lkuligin (approved):

PR #628, opened 2024-12-01
[vertexai] Add Hybrid Search Capabilities to VertexAI Vector Search

lspataroG:
🆕 **New Feature: Add Hybrid Search Capability to Vector Search**, see [here](https://cloud.google.com/vertex-ai/docs/vector-search/about-hybrid-search).
- Introduced function `add_texts_with_embeddings` to be able to pass both dense and sparse embeddings (useful anyway as many users already have precomputed embeddings).
-  Modified function `similarity_search_by_vector_with_score` to take in input a sparse embedding in addition to the dense one, added parameter `rrf_ranking_alpha` (default to 1) to be able to weight between dense and sparse importance during search.
- If a sparse embedding is passed then `similarity_search_by_vector_with_score` return both scores inside a dictionary with format: `[(Document, {"sparse_score": 0.2, "dense_score": 0.7})]`, otherwise the previous output format is preserved: `[(Document, 0.7)]`.
- Add tests: `test_add_texts_with_embeddings` and `test_similarity_search_by_vector_with_score_output_shape`


libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
```diff
@@ -30,17 +31,33 @@ class Searcher(ABC):
     def find_neighbors(
         self,
         embeddings: List[List[float]],
+        sparse_embeddings: Optional[List[Dict[str, List[int] | List[float]]]] = None,
```
lkuligin, line 34: can we make new arguments be explicitly called by name to ensure backwards compatibility, please?
so smth like `..., *, sparse_embeddings: ...`?

libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
```diff
@@ -30,17 +31,33 @@ class Searcher(ABC):
...
         k: int = 4,
+        rrf_ranking_alpha: float = 1,
         filter_: Union[List[Namespace], None] = None,
         numeric_filter: Union[List[NumericNamespace], None] = None,
-    ) -> List[List[Tuple[str, float]]]:
+    ) -> List[List[Dict[str, Any]]]:
```
lkuligin, line 39: same here, I'm a little bit concerned we're introducing a breaking change. Is there a chance we could keep it backwards compatible, please?

libs/vertexai/tests/unit_tests/test_vectorstores.py
```diff
@@ -76,3 +86,74 @@ def test_add_texts_with_custom_ids(mocker):
         VectorStore.add_texts(texts=texts, ids=["Id1"])
     with pytest.raises(ValueError):
         VectorStore.add_texts(texts=texts, ids=["Id1", "Id2", "Id2"])
+
+
+def test_add_texts_with_embeddings():
```
lkuligin, line 91: it's great we have a unit test! do you think it's worth adding an integration test too, please?

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
lspataroG: done

libs/vertexai/langchain_google_vertexai/vectorstores/_searcher.py
lspataroG: done

libs/vertexai/tests/unit_tests/test_vectorstores.py
lspataroG: added 2 integration tests:
- `test_vector_store_hybrid_search`
- `test_add_texts_with_embeddings`

lspataroG:
@lkuligin I am not sure why it is timing out

lkuligin (approved):

PR #629, opened 2024-12-04
[community] Fix BQ Vector Store batch search

lspataroG:
Fixed two bugs in BQ Vector Store inside the method `batch_search`
1. When passing queries instead of embeddings, the code was generating embeddings using `embedding_model.embed_documents` which is often the modality to encode documents instead of queries. 
2. The method works by creating a temporary table in BQ with embeddings in it and then performing the search using this table as input. When the batch is large enough however, the method `_bq_client.load_table_from_dataframe(df, table)` perform the operations in batches that results in saving a temp table where the order is not preserved. The fix consists in saving an additional field `row_num` that allow us to rebuild the initial input order.

🐛 Bug Fix


mariagpuyol (approved):
Thanks so much for this fix!

lkuligin (approved):

eliasecchig (approved):

PR #645, opened 2024-12-12
vertexai: Add context caching to VertexAI class

kardiff18:
## PR Description

- Add support for context caching to VertexAI class (currently it is only supported by ChatVertexAI)
- Refactor `is_gemini_advanced` into utils (to be used by both VertexAI and ChatVertexAI)

## Type

🆕 New Feature
🧹 Refactoring
✅ Test

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_base.py
+++ b/libs/vertexai/langchain_google_vertexai/_base.py
@@ -215,6 +215,11 @@ class _VertexAICommon(_VertexAIBase):
     model_name will be used to determine the model family
     """
 
+    cached_content: Optional[str] = None
+    """ Optional. Use the model in cache mode. Only supported in Gemini 1.5 and later
+        models. Must be a string containing the cache name (A sequence of numbers)
+    """
+
     @property
     def _is_gemini_model(self) -> bool:
         return is_gemini_model(self.model_family)  # type: ignore[arg-type]
--- a/libs/vertexai/langchain_google_vertexai/_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_utils.py
@@ -153,6 +153,10 @@ def is_gemini_model(model_family: GoogleModelFamily) -> bool:
     return model_family in [GoogleModelFamily.GEMINI, GoogleModelFamily.GEMINI_ADVANCED]
 
 
+def is_gemini_advanced(model_family: GoogleModelFamily) -> bool:
+    return model_family == GoogleModelFamily.GEMINI_ADVANCED
+
+
 def get_generation_info(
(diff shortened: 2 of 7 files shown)


libs/vertexai/langchain_google_vertexai/_base.py
```diff
@@ -215,6 +215,11 @@ class _VertexAICommon(_VertexAIBase):
     model_name will be used to determine the model family
     """
 
+    cached_content: Optional[str] = None
```
lkuligin, line 218: I'm sorry, but how would it be different from https://github.com/langchain-ai/langchain-google/blob/8c569edcab6b48f3f57390ded48f1af2e75eac8b/libs/vertexai/langchain_google_vertexai/chat_models.py#L1045?

kardiff18:
The issue today is that you cannot use `VertexAI` when using context caching, you can only use `ChatVertexAI`. If you try to use `VertexAI` with the cache it does not work.

So, the former PR that you linked it adds the parameter to the ChatVertexAI class, whereas this PR is adding it to VertexAI and therefore had to add it to the base model. Does that help?

lkuligin:
> The issue today is that you cannot use `VertexAI` when using context caching, you can only use `ChatVertexAI`. If you try to use `VertexAI` with the cache it does not work.
> 
> So, the former PR that you linked it adds the parameter to the ChatVertexAI class, whereas this PR is adding it to VertexAI and therefore had to add it to the base model. Does that help?

I'm wondering whether we should keep VertexAI at all. It's pretty outdated already and the gap will be increasing. Why do you need this class instead of ChatVertexAI?

kardiff18:
That's valid! My customer had attempted to use VertexAI at first since there was no multi-turn for the use case. After realizing context caching wasn't being properly used (there was no error message, we just knew the cache wasn't being invoked since the LLM response clearly was not using the cached instructions), I realized the issue was that they used VertexAI and not ChatVertexAI. I got them to switch over easily to ChatVertexAI after finding the issue.

That being said, I didn't know if anyone else was trying to use VertexAI without knowing it wasn't working properly due to a lack of error message, so I implemented it. We don't need this merge request, but I was just trying to help out others. We also mainly use ChatVertexAI for everything else. So in sum, I don't really think VertexAI is needed as well, so you're welcome to just ignore this PR if you think you'll be deprecating the class entirely.

lkuligin:
@kardiff18  to avoid a heavy code duplication, I wrapped VertexAI on top of ChatVertexAI here:

#749 

if you still feel there's a need for a context caching, let's do the same (just call ChatVertexAI and parse its output)

kardiff18:
That's fine! We can close this.

PR #646, opened 2024-12-13
vertexai: fix model family is always set even for non-google models

mjvankampen:
## PR Description

When using models outside the google family the model family is still set. This results in f.e. get_num_tokens hitting an enpoint while it should not.

## Relevant issues

N/A

## Type

🐛 Bug Fix

## Changes(optional)

N/A

## Testing(optional)

N/A

## Note(optional)

N/A


mjvankampen:
Nvm I think I need to use ChatAnthropicVertex

PR #660, opened 2024-12-22
vertexai: Add Intro to the Vertex AI library

SauravP97:
## PR Description

Added intro to the `Vertex AI` library.
Took reference from the Google Gen AI library: https://python.langchain.com/api_reference/google_genai/index.html

Rendered DocString looks like this: 
Screenshot 2024-12-22 at 3 13 21 PM

Tagging @lkuligin for review.

## Type

📖 Documentation


libs/vertexai/langchain_google_vertexai/__init__.py
```diff
@@ -1,3 +1,110 @@
...
+
+## Installation
+
+```bash
+pip install -U langchain-google-vertexai
+```
+
+## Chat Models
+
+`ChatVertexAI` class exposes models such as `gemini-pro` and `chat-bison`.
```
lkuligin, line 14: Palm is deprecated, not sure we should mention it.

libs/vertexai/langchain_google_vertexai/__init__.py
```diff
@@ -1,3 +1,110 @@
...
+
+This module contains the LangChain integrations for Google Cloud generative models.
+
+## Installation
+
+```bash
+pip install -U langchain-google-vertexai
+```
+
+## Chat Models
```
lkuligin, line 12: Let's mention also all models from Model Garden (MaaS and deployed on MG) and other integrations that are exposed in this package, please (like vector stores).

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/__init__.py
SauravP97: I don't see Palm in Model Garden supported Models. I can see Llama and Mistral only.
Please let me know if documentation for Palm needs to be added.

libs/vertexai/langchain_google_vertexai/__init__.py
SauravP97: Sure, I have added documentation for supported models in Model Garden i.e. Mistral and Llama.
I have also added documentation for the Vector Stores.

libs/vertexai/langchain_google_vertexai/__init__.py
```diff
@@ -1,3 +1,155 @@
...
+
+This module contains the LangChain integrations for Google Cloud generative models.
+
+## Installation
+
+```bash
+pip install -U langchain-google-vertexai
+```
+
+## Supported Models (MaaS: Model-as-a-Service)
```
lkuligin, line 12: let's start with a short description of all supported integration (and a difference between them, please). So that a user doesn't get an impression that only Llama and Mistarl are supported.
You can find a full list here: https://github.com/langchain-ai/langchain-google/blob/bd24d5097aa4a8a145b22fbca305652ded3df626/libs/vertexai/langchain_google_vertexai/__init__.py#L52 (we don't need to mention things like SafetySettings), but let's mention all relevant Vertex AI integrations, please.

SauravP97: Thanks for the review comment, looks like some integrated models were left in my PR. I have added those.

A short description of my recent commit.

1. Added Gemma models integration documentation
2. Added Vision models integration documentation
3. Added a difference between Chat model and Base model for the readers to understand our different flavors of models.

lkuligin (approved):

PR #658, opened 2024-12-19
genai: Fixed nested pydantic structures recursion

Shahar-Y:
## PR Description

While working with [gregpr07/browser-use](https://github.com/gregpr07/browser-use), @mahlertom and I found several bugs in langchain handling of nested `Pydantic` structured in `with_structured_output`. Specifically, the issues occurred in `_get_properties_from_schema` method in `libs\genai\langchain_google_genai\_function_utils.py`.

1. The current implementation did not recursively handle cases of nested `glm.Type.OBJECT` and only took care of the first layer.
2. In cases where `pydantic` fields are required (no default value is provided), the method did not properly add these fields to the `required` list.
3. In cases where an object of type `glm.Type.OBJECT` had empty properties, there was an error that properties of type `glm.Type.OBJECT` must not be empty.

## Relevant issues

Fixes https://github.com/langchain-ai/langchain-google/issues/657, 
May be related to https://github.com/langchain-ai/langchain/issues/24225

## Type 
🐛 Bug Fix

## Changes
As mentioned above:

1. The previous implementation did not recursively handle cases of nested `glm.Type.OBJECT` and only took care of the first layer. We fixed the recursion with:
```python
if properties_item.get("type_") == glm.Type.OBJECT:
    if v.get('anyOf') and isinstance(v['anyOf'], list) and isinstance(v['anyOf'][0], dict):
        v = v['anyOf'][0]
    v_properties = v.get("properties")
    if v_properties:
        properties_item["properties"] = _get_properties_from_schema_any(v_properties)
```

However, we found that when sending nested `pedantic` objects with required fields as expected output structure, like the case in the following method from [gregpr07/browser-use](https://github.com/gregpr07/browser-use) below:

```python
@time_execution_async('--get_next_action')
	async def get_next_action(self, input_messages: list[BaseMessage]) -> AgentOutput:
		"""Get next action from LLM based on current state"""

		structured_llm = self.llm.with_structured_output(self.AgentOutput, include_raw=True)
		response: dict[str, Any] = await structured_llm.ainvoke(input_messages)  # type: ignore

		parsed: AgentOutput = response['parsed']

		self._log_response(parsed)
		self.n_steps += 1

		return parsed
```

Here, when we observed the `response['parsed']` was often `None` and `response['parsing_error']` included validation errors for missing fields that were required.

2. To fix that error, we realized we needed to add the `required` list when parsing the `pydantic` objects. We added the nested `if` to our fix:

```python
[...]

PR #661, opened 2024-12-22
genai: Fix handling of optional arrays in tool input

piotr-rudnik:
## PR Description

This code fixes handling of optional arrays in pydantic models provided as tool inputs like this:
```   
class ExampleToolInput(BaseModel):
        numbers: Optional[List[str]] = Field()
```

Optional parameters are handled by setting `nullable` field in gemini API. Current code works for every type, except for arrays, because `items` field, that contains information about type of values inside the array, is present in `anyOf` field as one of the possibilities, where it's ignored.

Also added log where user is warned that only only type can be used inside array. Before, last defined type was used.
This is an edge case where user could be trying to use an array with multiple types which is not supported by genai, but can be useful for finding bugs

## Type

🐛 Bug Fix
✅ Test

## Testing(optional)

```
Launching pytest with arguments test_function_utils.py::test_tool_input_can_have_optional_arrays --no-header --no-summary -q in /xxx/langchain-google/libs/genai/tests/unit_tests

============================= test session starts ==============================
collecting ... collected 1 item

test_function_utils.py::test_tool_input_can_have_optional_arrays 

========================= 1 passed, 1 warning in 0.61s =========================
```

### DIFF
--- a/libs/genai/langchain_google_genai/_function_utils.py
+++ b/libs/genai/langchain_google_genai/_function_utils.py
@@ -314,6 +314,16 @@ def _get_properties_from_schema(schema: Dict) -> Dict[str, Any]:
 
         if properties_item.get("type_") == glm.Type.ARRAY and v.get("items"):
             properties_item["items"] = _get_items_from_schema_any(v.get("items"))
+        elif properties_item.get("type_") == glm.Type.ARRAY and v.get("anyOf"):
+            types_with_items = [t for t in v.get("anyOf") if t.get("items")]
+            if len(types_with_items) > 1:
+                len_types = len(types_with_items)
+                logger.warning(
+                    "Only first value for 'anyOf' key is supported in array types."
+                    f"Got {len_types} types, using first one: {types_with_items[0]}"
+                )
+            items = types_with_items[0]['items']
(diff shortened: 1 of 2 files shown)


Shahar-Y:
Hi, notice we already solved this bug in https://github.com/langchain-ai/langchain-google/pull/658 😄

piotr-rudnik:
Hey @Shahar-Y, I saw your PR before and it's similar but not the same, tried your code and didn't work in my case. Here, arrays are handled and they have `items` field setup, not `properties` in `v` object (I'm not that invested in the codebase, maybe it's not a problem). But those cases are very similar, maybe could be merged somehow.

lkuligin:
@piotr-rudnik could you rebase and fix linter errors, please?

piotr-rudnik:
@lkuligin Hey, branch got rebased, looks like all checks are passing.

lkuligin (approved):

piotr-rudnik:
There still were failing ruff checks, amended files to previous commit and force pushed new version

lkuligin:
@piotr-rudnik linter is still failing

krobert:
Hey, im getting the same error for ChatGroq too, i think this issue can be found in all basechatmodels

Issue #666, opened 2025-01-03
Make Google Drive API scopes configurable through GoogleDriveLoader cls


madhukar32:
## Issue Description
The GoogleDriveLoader currently hardcodes the Google Drive API scope to `["https://www.googleapis.com/auth/drive.file"]`. This restrictive scope only allows access to files created by the app itself, which doesn't work for many common use cases like reading existing files.

## Current Behavior

Scopes is hardcoded in the code and its not very easy to debug this issue, if someone is new to using Google API's
```python
SCOPES = ["https://www.googleapis.com/auth/drive.file"]
```
## Expected Behavior
Make scopes configurable via the GoogleDriveLoader class, with the current scope as the default:
```python
class GoogleDriveLoader(BaseLoader, BaseModel):
    scopes: List[str] = ["https://www.googleapis.com/auth/drive.file"]
```

## Example Use Case
I needed to read existing files and had to manually change the scope to:
pythonCopySCOPES = ["https://www.googleapis.com/auth/drive.readonly"]
This took significant debugging time to identify as the root cause of access issues.

lkuligin:
I believe we can close it (thanks to your contribution), can't we?

madhukar32:
> I believe we can close it (thanks to your contribution), can't we?

Yes we can, thanks for pro-actively reviewing this

PR #667, opened 2025-01-03
community: add scope validation for Google Drive API credentials

madhukar32:
## PR Description

This PR makes the Google Drive API scopes configurable in the GoogleDriveLoader class instead of using a hardcoded global constant. This change allows users to specify different scopes based on their needs while maintaining backward compatibility with the default `drive.file` scope.

Currently, the scope is hardcoded to `drive.file` which is restrictive and only allows access to files created by the app. This change enables users to use different scopes like `drive.readonly` for broader read access without modifying the source code.

## Relevant issues

ref: https://github.com/langchain-ai/langchain-google/issues/666

## Type

🐛 Bug Fix

## Changes(optional)

- Added configurable `scopes` parameter to `GoogleDriveLoader` class
- Implemented scope validation with predefined valid scopes
- Maintained backward compatibility by using original scope as default
- Added documentation for available scopes
- Added unittest to verify the scopes

### DIFF
--- a/libs/community/langchain_google_community/drive.py
+++ b/libs/community/langchain_google_community/drive.py
@@ -9,17 +9,25 @@
 
 import os
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Sequence, Union
+from typing import Any, ClassVar, Dict, List, Optional, Sequence, Tuple, Union
 
 from langchain_core.document_loaders import BaseLoader
 from langchain_core.documents import Document
 from pydantic import BaseModel, field_validator, model_validator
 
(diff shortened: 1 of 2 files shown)


lkuligin (approved):

lkuligin:
@madhukar32 Thanks for your contribution! Please, take a look at remaining linter errors, and I'm happy to merge when it's fixed.

madhukar32:
> @madhukar32 Thanks for your contribution! Please, take a look at remaining linter errors, and I'm happy to merge when it's fixed.

Thanks @lkuligin for your response, I fixed the lint errors and pushed it again.

PR #671, opened 2025-01-03
genai: Fix multiple tool calls in a single AIMessage

pedrito87:
## PR Description

This PR fixes known issues with Langchain Google genai, where for multiple tool calls in a single AI Message there was an error from Google generative ai library, stating that the number of parts connected to function calls must be equal the number of function responses.

## Relevant issues

["Fixes #27799"](https://github.com/langchain-ai/langchain/issues/27799)
This issue is mentioned in a couple of places, without identifying the real issue behind this behaviour. This PR fixes the underlying logic problem in parsing the chat history.

## Type

🐛 Bug Fix

## Changes(optional)

The `_parse_chat_history` function gets a fix for handling multiple tool calls in a single AIMessage.

## Testing(optional)

The relevant unit tests are updated.
In order to check the fix you can check the behaviour of the library before and after the fix, making the model call 2 tools in a single AIMessage.

PR #681, opened 2025-01-08
Fix integration test with multiple parts

pedrito87:
## PR Description

Fix integration test connected to the [multi part tool calling](https://github.com/langchain-ai/langchain-google/pull/671).

## Relevant issues

This is connected to this [fix](https://github.com/langchain-ai/langchain-google/pull/671) from previous PR.

## Type

🐛 Bug Fix
✅ Test

## Changes(optional)

Integration test `test_chat_function_calling_with_multiple_parts` can now be run without failing.


pedrito87:
@lkuligin you can take a look at this integration test fix. It is possible for it not to fail now thanks to my [previous PR](https://github.com/langchain-ai/langchain-google/pull/671). Cheers.

lkuligin:
integration tests are green at the moment, are you sure we need this PR?

pedrito87:
> integration tests are green at the moment, are you sure we need this PR?

@lkuligin They are green because this test is marked as [pytest `xfail`](https://docs.pytest.org/en/stable/how-to/skipping.html), which runs the test and the marks the test as XFAIL (if it actually fails) or XPASS (if it passes despite being marked as xfail). Both cases are considered a success in terms of CI/CD.

My change actually fixes this integration test, which in the current form has no chance of succeeding, and is a good example of how the system works in terms of multiple tools uses.

And when you compare the integration test between the main branch and this one you will see that the test fixed here is actually an XPASS on this branch, compared to XFAIL on main branch.

So, answering your question, I think it is worth merging 👍

lkuligin (approved):

PR #693, opened 2025-01-13
removed a limitation on pydantic version

lkuligin:


ostefano:
@lkuligin note that this bump is also necessary: google-cloud-aiplatform>=1.76.0

PR #695, opened 2025-01-14
vertexai[patch]: support Codestral 25.01 in model garden maas

ccurme:


lkuligin:
enabled the model for the test project, now tests should pass

lkuligin (approved):

PR #701, opened 2025-01-17
Add support for structured output to VertexAI class

izo0x90:
## PR Title
Add support for structured output to VertexAI class

## PR Description

Having guaranteed json or other supported structured output can be very useful for the chat use-cases as well, updating the VertexAI to support setting a schema in the same way that the Chat model class implements it.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/llms.py
+++ b/libs/vertexai/langchain_google_vertexai/llms.py
@@ -16,6 +16,9 @@ from vertexai.generative_models import (  # type: ignore[import-untyped]
     GenerativeModel,
     Image,
 )
+from vertexai.generative_models._generative_models import (  # type: ignore[import-untyped]
+    _convert_schema_dict_to_gapic,
+)
 from vertexai.language_models import (  # type: ignore[import-untyped]
     CodeGenerationModel,
     TextGenerationModel,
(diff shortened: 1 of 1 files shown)


lkuligin:
in general, I'm more and more wondering whether we should continue supporting VertexAI or mark it as deprecated
it's also related to #645 

https://python.langchain.com/docs/concepts/text_llms/
https://python.langchain.com/docs/integrations/llms/ 

@efriis @baskaryan @ccurme any additional thoughts?

izo0x90:
> in general, I'm more and more wondering whether we should continue supporting VertexAI or mark it as deprecated it's also related to #645
> 
> https://python.langchain.com/docs/concepts/text_llms/ https://python.langchain.com/docs/integrations/llms/
> 
> @efriis @baskaryan @ccurme any additional thoughts?

@lkuligin 
For this would the idea be to use chat class for use cases that do not fit the "Chat" modality? Is the behavior there essentially the same as the VertexAI call if we don't pass in message history?

lkuligin:
> @lkuligin For this would the idea be to use chat class for use cases that do not fit the "Chat" modality? Is the behavior there essentially the same as the VertexAI call if we don't pass in message history?

yes, you essentially (chat_llm | StrOutputParser()).invoke(text_input) should have the same effect. Let me know if you have any different use case, please.

izo0x90:
@lkuligin https://github.com/langchain-ai/langchain-google/pull/763 <<< here is a PR that that implements passing through the schema confing to the ChatVertexAI client.

PR #714, opened 2025-01-26
Added support for max_tokens attribute in ChatVertexAI

### LINKED ISSUE #596: ChatVertexAI has no property max_tokens.
ChatVertexAI inherits from BaseChatModel which accepts a standard property, `max_tokens`, from the documentation. 

When initializing a ChatVertexAI instance with this standard property, it is not available on the instance and throws an error upon accessing. It is instead available under `max_output_tokens`. 

It should be made available on the instance as the other chat model integrations (ChatOpenAI, ChatAnthropic, ...) all have this property available on the instance.

```python
model = ChatVertexAI(
    model="gemini-1.5-pro-002",
    temperature=0.0,
    max_tokens=500,
    top_p=1.0,
)

print(model.max_output_tokens)  # 500
print(model.max_tokens).  # AttributeError: 'ChatVertexAI' object has no attribute 'max_tokens'
```

---
LangChain package versions:

langchain-core==0.3.15
langchain-google-vertexai==2.0.7

SauravP97:
## PR Description

Added support for `max_tokens` attribute in `ChatVertexAI`

## Relevant issues

Fixes #596 

## Type

🐛 Bug Fix
✅ Test

Tagging @lkuligin for a review.


libs/vertexai/langchain_google_vertexai/_base.py
```diff
@@ -182,6 +182,7 @@ class _VertexAICommon(_VertexAIBase):
     temperature: Optional[float] = None
     "Sampling temperature, it controls the degree of randomness in token selection."
     max_output_tokens: Optional[int] = Field(default=None, alias="max_tokens")
+    max_tokens: Optional[int] = Field(default=None, alias="max_tokens")
```
lkuligin, line 185: I'd rather make it a @property. I don't think we need to store an extra field.

SauravP97: Thanks for pointing out. I believe that's why the unit test was failing. Added this as a property.

PR #715, opened 2025-01-26
genai: fix function_utils to parse object/array json schemas properly

RazCrimson:
The existing code tries to handle arrays and objects in anyOf fields of JSON Schema's for tools by picking the last possible non nullable choice in anyOf. But the context of the field is not updated to reflect the selected choice leading to a borked conversion to protobuf that fails with 400 errors.

This fix is just to update the context for arrays and objects.

Please correct me if it is intentional, but there is already a testcase which is supposed to cover this, but it is buggy as it sets an `items` field within an `items` definition for an array, which should not be possible according to the JSON Schema spec. The faulty setting made this bug slip through the tests


lkuligin:
thanks for the change, please, fix linting errors.

I'm not sure it's a bug, we probably have it intentianally since old models didn't support `AnyOf` fields. But anyways, it shouldn't prevent us from fixing this.

RazCrimson:
@lkuligin Fixed lint errors

PTAL

lkuligin (approved):

PR #716, opened 2025-01-27
vertexai: Add `embeddings_task_type` parameter to `embed_query` and `embed_documents`

mitraan-deshaw:
## PR Description
This PR introduces an optional `embeddings_task_type` parameter to `embed_documents` and `embed_query` in `VertexAIEmbeddings`, mirroring the `task_type` argument in [GoogleGenerativeAIEmbeddings](https://github.com/langchain-ai/langchain-google/blob/106872f37ccbe8329a38a68293e4847cf0f509b2/libs/genai/langchain_google_genai/embeddings.py#L187). While working on this, I also addressed a couple of additional improvements:
- Created `EmbeddingTaskTypes` to consolidate embedding task types, reducing redundancy. I've currently added it as a Literal type but I'm open to suggestions for a more suitable location.
- Fixed an issue where the `text-embedding-005` model, which supports embedding tasks, was incorrectly classified as an older model, preventing proper argument passing. I've added a new `GoogleEmbeddingModelVersion` to address this.
- Added (new) `CODE_RETRIEVAL_QUERY` as an embedding task. 

Sample code:
```python
from langchain_google_vertexai import VertexAIEmbeddings

# Initialize the a specific Embeddings Model version
embeddings = VertexAIEmbeddings(model_name="text-embedding-005")
single_vector = embeddings.embed_query(text, embeddings_task_type="CODE_RETRIEVAL_QUERY")
```

## Type

🆕 New Feature

## Changes(optional)

- Added optional `embeddings_task_type` parameter to `embed_documents` and `embed_query` in `VertexAIEmbeddings`.
- Consolidated embeddings task types into `EmbeddingTaskTypes` to reduce redundancy.
- Added support for `text-embedding-005` embedding tasks. 
- Added `CODE_RETRIEVAL_QUERY` as an embedding task.


libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -447,7 +454,10 @@ def embed(
         return embeddings
 
     def embed_documents(
-        self, texts: List[str], batch_size: int = 0
+        self,
+        texts: List[str],
+        batch_size: int = 0,
+        embeddings_task_type: EmbeddingTaskTypes = "RETRIEVAL_DOCUMENT",
```
lkuligin, line 460: please, add a `*,` so that `embeddings_task_type` can be provided by name only

libs/vertexai/tests/integration_tests/test_embeddings.py
```diff
@@ -65,6 +95,26 @@ def test_langchain_google_vertexai_embedding_query(model_name, embeddings_dim) -
...
+@pytest.mark.release
+@pytest.mark.parametrize(
+    "embeddings_task_type",
+    _EMBEDDING_TASK_TYPES,
+)
+@pytest.mark.parametrize(
+    "model_name, embeddings_dim",
+    _EMBEDDING_MODELS,
+)
+def test_langchain_google_vertexai_embedding_query_with_task_type(
```
lkuligin, line 107: do we need this integration test, or would a unit test be enough (we've tested in a test above that embeddings_task is passed to the Google API and it returns a valid output)

we've got too many integration tests now and the execution time gets longer and longer :), it might be a good idea to keep the total amount of them reasonable

lkuligin (changes requested):

libs/vertexai/tests/integration_tests/test_embeddings.py
mitraan-deshaw: Totally makes sense! However, I'm not sure how we would write a clean unit test for the new argument in `embed_documents()` and `embed_query()`, as these methods in turn call `embed()` and return its response. It almost seems like we would be validating the `embed()` method instead.

If we only want to check if the arguments are passed correctly, one approach could be to mock the `embed()` method, capture the arguments, and then call the `VertexAIEmbedding`'s `embed()` method. Happy to remove the tests if you think it's not required.

lkuligin: just by mocking the API: https://github.com/langchain-ai/langchain-google/blob/c3d40615a1c60baf961a4fb5d745b51b6032da72/libs/vertexai/tests/unit_tests/test_embeddings.py#L20

libs/vertexai/langchain_google_vertexai/embeddings.py
mitraan-deshaw: fixed

libs/vertexai/tests/integration_tests/test_embeddings.py
mitraan-deshaw: Moved the tests into unit tests.

lkuligin (approved):

lkuligin:
@mitraan-deshaw could you fix the linter, please?

lkuligin (approved):

PR #722, opened 2025-01-31
Ensure a single instance of ImageBytesLoader throughout chat_models

### LINKED ISSUE #626: The project is still crating many gcs clients
After i fixed: https://github.com/langchain-ai/langchain-google/commit/ef446127190951ebb89a55914c7e489423534639

I suddenly saw that ImageBytesLoader is created ad hoc in every function. So the commit mentioned is just a first step to fix this issue.

Look please here: https://github.com/search?q=repo%3Alangchain-ai%2Flangchain-google%20ImageBytesLoader&type=code

What do you suggest here?

The path which i hit is at https://github.com/langchain-ai/langchain-google/blob/7f704aad855091e8e3be834368a3e08571d7cce1/libs/vertexai/langchain_google_vertexai/chat_models.py#L221 and this is a stateless function.

I see  one possible solution which is creating a cache of imagebytsloader. for xample creating a map from project to imagbytesloader (lru_cache) . But this is a bigger change and want to hear what you think.

SauravP97:
## PR Description

Ensuring a single instance of ImageBytesLoader class being created throughout chat_models. This will further make sure a single Google Cloud Storage client is being used throughout.

## Relevant issues

Fixed #626 

## Type

🐛 Bug Fix
✅ Test

Adding @lkuligin for a review.


libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -1076,6 +1077,10 @@ class Joke(BaseModel):
         models. Must be a string containing the cache name (A sequence of numbers)
     """
 
+    project: Optional[str] = None
```
lkuligin, line 1080: we already have a project_id from the base class, why do we need a separate one?

lkuligin (changes requested):

lkuligin (approved):

PR #735, opened 2025-02-14
docs: add temp_dataset_name to community.BigQueryVectorStore docs

cmenon12:
## PR Description

Added `temp_dataset_name` to the docstring for `BigQueryVectorStore`, and an explanation about how to use it to avoid needing permission to create datasets in GCP.

This attribute is defined in `BaseBigQueryVectorStore`, albeit without the additional explanation.

## Relevant issues
None.

## Type

📖 Documentation

## Note(optional)

In my environment I'm not allowed to create datasets, so was confused why `BigQueryVectorStore` asked for the `bigquery.datasets.create` permission even though the dataset already existed. Digging into the source code I found the `temp_dataset_name` attribute, and specifying it as the name of a dataset that already existed stopped this class from asking for this permission.

### DIFF
--- a/libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py
+++ b/libs/community/langchain_google_community/bq_storage_vectorstores/bigquery.py
@@ -36,6 +36,9 @@ class BigQueryVectorStore(BaseBigQueryVectorStore):
     `batch_search` method.
     Optionally, this class can leverage a Vertex AI Feature Store for online serving
     through the `to_vertex_fs_vector_store` method.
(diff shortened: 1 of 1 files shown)


cmenon12:
Hi @lkuligin this is ready for review. I see the integration tests have failed, although I would expect no change given that only the docs have changed. Please let me know if I need to recreate the PR.

lkuligin:
@cmenon12 thanks for your contribution!
please, fix the linter and I'm happy to merge it.

cmenon12:
@lkuligin thank you very much! I think I've fixed the issue the linter found.

PR #761, opened 2025-02-26
genai[patch]: update gemini 1.0 -> 1.5 and 2.0 in tests

ccurme:
Tests failing with
> google.api_core.exceptions.NotFound: 404 models/gemini-1.0-pro-001 is not found for API version v1beta, or is not supported for generateContent. Call ListModels to see the list of available models and their supported methods.

(Unsure why but model is scheduled to be retired in 1-2 months anyway.)


tarikaltuncu:
Hey!

PyPI package is not updated with this fix, and many applications are failing with `google.api_core.exceptions.NotFound: 404 models/gemini-1.0-pro is not found for API version v1beta, or is not supported for generateContent. Call ListModels to see the list of available models and their supported methods.` because of this. 

How do we trigger a PyPI update with this patch?

PR #766, opened 2025-03-06
Bug Fix: Make call of literal_eval optional

### LINKED ISSUE #710: Usage of `literal_eval` considered unsafe
The chat models currently use [`literal_eval`](https://github.com/langchain-ai/langchain-google/blob/1e1a1b7cf688d4889af180c054b8ab1d279b0fec/libs/vertexai/langchain_google_vertexai/chat_models.py#L258-L264) to parse messages. 

The [official Python documentation](https://docs.python.org/3/library/ast.html#ast.literal_eval) states:
> Warning It is possible to crash the Python interpreter due to stack depth limitations in Python’s AST compiler.

Also [official CPython contributor vstinner states](https://github.com/python/cpython/issues/95588#issuecomment-1266947505):
> It was never safe.

Since messages clearly contain user input, we currently consider this an issue which prevents us from using this implementation in production. Are there any plans to rework this part of the code and eliminate the `literal_eval`?

hzeus:
## PR Description

Add a flag to make the call of `literal_eval` optional

## Relevant issues

Fixes #710 

## Type

🐛 Bug Fix

## Note(optional)

Hi 👋 

This is my attempt to make the usage of `literal_eval` optional. I added a flag to the ChatModel and set the default to `True` to keep it backwards compatible - however out of a security perspective I would rather suggest setting it to `False`.

In our case (we do not use `ChatPromptTemplate`), the usage of `literal_eval` is not necessary, but even leads to a bug when entering a value that is parseable as a Float ("3.14") into our chat (`TypeError: 'float' object is not iterable`)

### DIFF
--- a/libs/vertexai/tests/integration_tests/test_chat_models.py
+++ b/libs/vertexai/tests/integration_tests/test_chat_models.py
@@ -344,7 +344,9 @@ def test_parse_history_gemini_multimodal_FC():
     expected = [Content(role="user", parts=parts)]
     imageBytesLoader = ImageBytesLoader()
     _, response = _parse_chat_history_gemini(
-        history=history, imageBytesLoader=imageBytesLoader
+        history=history,
+        imageBytesLoader=imageBytesLoader,
+        perform_literal_eval_on_string_raw_content=True,
     )
     assert expected == response
 
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -203,6 +203,7 @@ def _parse_chat_history_gemini(
     history: List[BaseMessage],
     imageBytesLoader: ImageBytesLoader,
     convert_system_message_to_human: Optional[bool] = False,
+    perform_literal_eval_on_string_raw_content: Optional[bool] = False,
 ) -> tuple[Content | None, list[Content]]:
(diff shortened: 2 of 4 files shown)


libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -344,7 +344,9 @@ def test_parse_history_gemini_multimodal_FC():
     expected = [Content(role="user", parts=parts)]
     imageBytesLoader = ImageBytesLoader()
     _, response = _parse_chat_history_gemini(
-        history=history, imageBytesLoader=imageBytesLoader
+        history=history,
+        imageBytesLoader=imageBytesLoader,
+        perform_literal_eval_on_string_raw_content=True,
```
lkuligin, line 349: nits: the default is True anyways, do we really need it here?

maybe a unit test would be nice that tests the desired behaviour

lkuligin (approved):

hzeus:
Thanks for your review!

I added two unit tests for `_parse_chat_history_gemini` and also adjusted the snapshot which should fix the test runs.

> nits: the default is True anyways, do we really need it here?

I set the default to `True` to the initialization of the Model, but not for the direct call of the function.

lkuligin (approved):

hzeus:
I just noticed that the previous commit failed the linter. Hopefully fixed that.

PR #772, opened 2025-03-11
genai[patch]: update langchain-tests, fix bugs with structured output

ccurme:
- Fix output parser in case when schema is specified via TypedDict or JSON Schema
- Update `langchain-tests` and support tracing structured output


libs/genai/tests/integration_tests/test_chat_models.py
```diff
@@ -457,15 +457,7 @@ class MyModel(BaseModel):
...
-    expected = [
-        {
-            "type": "MyModel",
-            "args": {
-                "name": "Erick",
-                "age": 27,
-            },
-        }
-    ]
+    expected = {"name": "Erick", "age": 27}
```
ccurme, line 460: This is a breaking change. The previous behavior was not consistent with other LangChain chat models (including ChatVertexAI).

lkuligin (approved):

Issue #691, opened 2025-01-13
400 Bad Request if sending messages with empty content (contents.parts must not be empty).


splitpierre:
ERROR [Error: [GoogleGenerativeAI Error]: Error fetching from https://generativelanguage.googleapis.com/v1beta/models/gemini-2.0-flash-exp:streamGenerateContent?alt=sse: [400 Bad Request] * GenerateContentRequest.contents[41].parts: contents.parts must not be empty.

While implementing chat with history, be careful in cases where some Message property "content" comes empty, it will cause this error. Apparently this is only happening with ChatGoogleGenerativeAI, I'm using several others and didn't had this issue.
For now i made sure to filter those empty entries out, but still, i think this should be handled either by Google or by this repo.

Thanks!

lkuligin:
could you add more details on how do you get a message with empty content, please? what's your use case?

PR #775, opened 2025-03-12
genai : filter out empty message contents in ChatGoogleGenerativeAI

asimchoudhary:
## PR Description
It  filters out the messages with content parameter being empty  in the `_prepare_request` method.

## Relevant issues

Fixes : #691

## Type

🐛 Bug Fix

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -1177,7 +1177,15 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
         elif functions:
             formatted_tools = [convert_to_genai_function_declarations(functions)]
 
-        messages = [message for message in messages if message.content]
+        filtered_messages = []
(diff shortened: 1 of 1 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1177,6 +1177,8 @@ def _prepare_request(
         elif functions:
             formatted_tools = [convert_to_genai_function_declarations(functions)]
 
+        messages = [message for message in messages if message.content]
```
lkuligin, line 1180: a message can have an empty content but `tool_calls`.

probably, only HumanMessages should be filtered out (and please raise a warning!)

lkuligin (changes requested):

libs/genai/langchain_google_genai/chat_models.py
asimchoudhary: On it !

lkuligin (approved):

PR #777, opened 2025-03-12
Avoid retries for InvalidArgument exceptions

### LINKED ISSUE #762: Stop retrying when got InvalidArgument exception
The following code aims to ask gemini to describe an image by giving the image uri.
I am testing on an invalid image uri, and google raises InvalidArgument exception.

I suppose in this case, langchain should not retry when the input is invalid.

Is there a way to stop retry for this case but keep retrying for other exceptions, e.g. ResourceExhausted and ServiceUnavailable?

If not, could you please implement the feature to treat this exception differently?

Following is the code I used:

```python

from loguru import logger

load_dotenv(override=True)

llm = ChatVertexAI(
        model="gemini-1.5-flash",
        location="europe-west1",
        max_output_tokens=8092,
        temperature=0,
        max_retries=10,
    )

image_uri = "https://upload.wikimedia.org/wikipedia/commons/3/38/JPEG_xample_JPG_RIP_001.jpg" #invalid image uri

media_message = {
    "type": "media",
    "file_uri": image_uri,
    "mime_type": "image/jpeg",
}
text_message = {
    "type": "text",
    "text": "Describe the image.",
}
message = HumanMessage(content=[text_message, media_message])

response = llm.invoke([message])
logger.debug(response)
```

SauravP97:
Avoid retries for `InvalidArgument` exception.

Currently `GoogleAPIError` has been allowed for retries. [`InvalidArgument`](https://googleapis.dev/python/google-api-core/latest/exceptions.html#google.api_core.exceptions.InvalidArgument) is a child-class  of `GoogleAPIError` and hence being included in the retries.

## Relevant issues

Fixes #762 

## Type

🐛 Bug Fix
🧹 Refactoring
✅ Test

## Testing(optional)

1. Tested for successful retries in case of `GoogleAPIError` exception and their sub-class exceptions.
2. Tested for un-successful retries and immediate exceptions being throws for `InvalidArgument` type.

## Note(optional)

Parent-child relationship.

parent-child-exception

Adding @lkuligin for a review.


libs/vertexai/langchain_google_vertexai/utils.py
```diff
@@ -148,9 +153,24 @@ def _before_sleep(retry_state: RetryCallState) -> None:
     max_seconds = 10
     # Wait 2^x * 1 second between each retry starting with
     # 4 seconds, then up to 10 seconds, then 10 seconds afterwards
-    retry_instance: retry_base = retry_if_exception_type(error_types[0])
+
+    google_api_call_error_retry_instance = retry_if_exception_type(
```
lkuligin, line 157: afaik, it's easier first to check whether `GoogleAPICallError in error_types` and then construct a retry_instance accordingly

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/utils.py
SauravP97: Sure makes sense, done

lkuligin (approved):

PR #780, opened 2025-03-16
fix[genai]: Allow Temperature Values Up to 2.0 for ChatGoogleGenerativeAI

### LINKED ISSUE #781: BUG[genai}: Temperature Validation Restriction Causes Incompatibility with Gemini API
The current temperature validation logic in the `ChatGoogleGenerativeAI` module is limited to a maximum of **1.0**. However, the Gemini API supports values up to **2.0**. As a result, when users try to set a temperature value higher than 1.0, the validation fails—forcing workarounds or leading to test failures. This issue restricts the creative potential that the Gemini API can offer.

#### Repro Steps

**To Reproduce:**

1. Set the temperature value to something greater than **1.0**.
2. Initialize the `ChatGoogleGenerativeAI` model with the elevated temperature.
3. Observe that the validation error is raised.

**Example Code:**

```python
import os

os.environ["GOOGLE_API_KEY"] = "..."

# Attempting to use a temperature greater than 1.0 causes a validation error
model = ChatGoogleGenerativeAI(model="gemini-2.0-flash", temperature=1.5)
print(model.invoke("Hello, world!"))
```

#### Expected Behavior

The temperature validation should accept values in the range **[0.0, 2.0]**. This will allow users to leverage the full creative range supported by the Gemini API without encountering validation errors.

mjunaidca:
## PR Description

Current temperature validation restricts values to a maximum of 1.0. However, the Gemini API supports values up to 2.0, which allows for more creative and diverse outputs. This discrepancy led to failing code and forces to users to directly use gemini api when we need temperature value >1.

### Changes Made
- Validation Update: Modified the temperature validation logic to accept values within the range [0.0, 2.0] instead of the previous [0.0, 1.0].
- Test Enhancement: Added a new unit test (test_temperature_range) to verify that temperature values above 1.0 up to 2.0 are accepted.
- Compatibility: Ensured that this change aligns our implementation with the Gemini API's supported range, enabling more flexible usage without breaking existing functionality.

## Relevant issues

Fixes #781

## Type

🐛 Bug Fix

## Changes(optional)

- Updated temperature validation to allow values in range [0.0, 2.0] instead of [0.0, 1.0]
- This ensures compatibility with Gemini API's supported temperature range
- Create a new Unit Test that is now passing to check for temperature values in this range

## Testing(optional)

- Added a new test test_temperature_range that now passes correctly
- Verified that temperature values between 0 and 2.0 are properly accepted

## Note(optional)
This change aligns our code with the Gemini API's capabilities, allowing users to set higher temperature values when needed for more creative outputs.

- [x] No Breaking Changes @lkuligin @SauravP97 
- [x] Necessary tests have been added
- [x] Lint and tests pass successfully

### DIFF
--- a/libs/genai/tests/unit_tests/test_chat_models.py
+++ b/libs/genai/tests/unit_tests/test_chat_models.py
@@ -646,7 +646,7 @@ def test__convert_tool_message_to_part__sets_tool_name(
     assert part.function_response.response == {"output": "test_content"}
 
 
-def test_temperature_range() -> None:
+def test_temperature_range_pydantic_validation() -> None:
     """Test that temperature is in the range [0.0, 2.0]"""
 
     with pytest.raises(ValidationError):
@@ -669,7 +669,7 @@ def test_temperature_range() -> None:
     }
 
 
-def test_temperature_range_value_error() -> None:
+def test_temperature_range_model_validation() -> None:
     """Test that temperature is in the range [0.0, 2.0]"""
 
     with pytest.raises(ValueError):
--- a/libs/genai/tests/unit_tests/test_chat_models.py
+++ b/libs/genai/tests/unit_tests/test_chat_models.py
@@ -24,6 +24,7 @@ from langchain_core.messages import (
 )
 from langchain_core.messages.tool import tool_call as create_tool_call
 from pydantic import SecretStr
+from pydantic_core._pydantic_core import ValidationError
 from pytest import CaptureFixture
 
 from langchain_google_genai.chat_models import (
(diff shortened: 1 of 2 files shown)


libs/genai/tests/unit_tests/test_chat_models.py
```diff
@@ -643,3 +644,26 @@ def test__convert_tool_message_to_part__sets_tool_name(
     part = _convert_tool_message_to_part(tool_message)
     assert part.function_response.name == "tool_name"
     assert part.function_response.response == {"output": "test_content"}
+
+
+def test_temperature_range() -> None:
+    """Test that temperature is in the range [0.0, 2.0]"""
+
+    with pytest.raises(ValidationError):
```
SauravP97, line 652: Can we also assert for the `ValueError` message being thrown in the SUT?

```
raise ValueError("temperature must be in the range [0.0, 2.0]")
```

SauravP97 (changes requested):

libs/genai/tests/unit_tests/test_chat_models.py
mjunaidca: @SauravP97 I've updated the tests to better handle both cases and assert the expected error messages.

lkuligin (approved):

Issue #730, opened 2025-02-10
Support Code Execution for Gemini models


ryanjhoffman:
Gemini 1.5 pro and Gemini 2.0 flash both support code execution via the google genai API: 

https://ai.google.dev/gemini-api/docs/code-execution?lang=python

From what I can tell, there is not currently a way to enable this feature when using langchain.

asimchoudhary:
On it !

PR #788, opened 2025-03-17
(genai): Add support for Gemini code execution

asimchoudhary:
Fixes : #730

This PR adds support for the code execution feature available in Gemini models (1.5 Pro, 1.5 Flash, 2.0 Flash, and 2.0 Pro).

### Current Implementation Status:
- [x] Added `code_execution` boolean parameter to `ChatGoogleGenerativeAI` (default: False)
- [x] Added model validation to ensure code execution is only enabled on supported models
- [x] Updated `_prepare_request` to add a `CodeExecution` tool when enabled
- [ ] Update `_parse_response_candidate` to handle code execution results
- [ ] Add tests for code execution
- [ ] Add documentation and examples

#### @lkuligin, I’d appreciate your feedback on the following:
1. Is this the right approach for implementing code execution support?
2. For parsing responses with code execution results, I'm planning to modify `_parse_response_candidate` to handle:
   - `executable_code` parts
   - `code_execution_result` parts
   - Inline data (images/charts)
   
   Is there a preferred way to structure this in LangChain's message format?
`Current Output from Local Testing:
Here’s how the implementation behaves when executed locally:`
Screenshot from 2025-03-17 15-30-33

🆕 New Feature

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -75,7 +75,7 @@ from langchain_core.output_parsers.openai_tools import (
     parse_tool_calls,
 )
 from langchain_core.outputs import ChatGeneration, ChatGenerationChunk, ChatResult
-from langchain_core.runnables import Runnable, RunnablePassthrough
+from langchain_core.runnables import Runnable, RunnablePassthrough, RunnableConfig
 from langchain_core.tools import BaseTool
 from langchain_core.utils.function_calling import convert_to_openai_tool
 from pydantic import (
@@ -811,12 +811,7 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
     (e.g. what content to cache) and enjoy guaranteed cost savings. Format: 
     ``cachedContents/{cachedContent}``.
     """
-    code_execution: bool = False
-    """Whether to enable code execution capabilities.
-    
-    Only supported on models: gemini-1.5-pro, gemini-1.5-flash, gemini-2.0-flash, and gemini-2.0-pro.
-    When enabled, the model can execute code to solve problems.
-    """
+
 
     model_config = ConfigDict(
         populate_by_name=True,
@@ -831,7 +826,7 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
         return "chat-google-generative-ai"
     
     @property
-    def _supported_code_execution(self)->bool:
+    def _supports_code_execution(self)->bool:
         return(
             "gemini-1.5-pro" in self.model
             or "gemini-1.5-flash" in self.model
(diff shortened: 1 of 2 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -820,6 +829,15 @@ def lc_secrets(self) -> Dict[str, str]:
     @property
     def _llm_type(self) -> str:
         return "chat-google-generative-ai"
+    
+    @property
+    def _supported_code_execution(self)->bool:
```
lkuligin, line 834: nits: rename to `_supports_code_execution`, please

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1171,11 +1189,20 @@ def _prepare_request(
                 "Must specify at most one of tool_choice and tool_config, received "
                 f"both:\n\n{tool_choice=}\n\n{tool_config=}"
             )
+              
         formatted_tools = None
         if tools:
             formatted_tools = [convert_to_genai_function_declarations(tools)]
         elif functions:
             formatted_tools = [convert_to_genai_function_declarations(functions)]
+        elif self.code_execution:
```
lkuligin, line 1198: let's allow to pass `code_execution=True` as part of kwargs of the `invoke` method, please

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1171,11 +1189,20 @@ def _prepare_request(
                 "Must specify at most one of tool_choice and tool_config, received "
                 f"both:\n\n{tool_choice=}\n\n{tool_config=}"
             )
+              
         formatted_tools = None
         if tools:
```
lkuligin, line 1194: we should raise a warning (or even an error?) if we have both `tools` and `code_execution=True`

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -808,6 +811,12 @@ class Joke(BaseModel):
     (e.g. what content to cache) and enjoy guaranteed cost savings. Format: 
     ``cachedContents/{cachedContent}``.
     """
+    code_execution: bool = False
```
lkuligin, line 814: I doubt we need this as part of the instance's argument, I'd rather pass it only as part of `invoke` kwargs. Do you envision the use case when you want to use an instance of a model **only** with the code execution tool?

And anyways, you can always achieve it with `llm.bind(code_execution=True)`

lkuligin:
@asimchoudhary thanks for your contribution!

let's follow approach similar to reasoning models - if you have a code_execution results, let's transform it to a dict and make part of a multi-part message.content. WDYT?

asimchoudhary:
@lkuligin  Thanks for your feedback! I’ll implement the suggested changes.  
Regarding the parsing of `code_execution` output, I completely agree with using the reasoning model approach.  

On it!

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1197,12 +1293,18 @@ def _prepare_request(
                 "Must specify at most one of tool_choice and tool_config, received "
                 f"both:\n\n{tool_choice=}\n\n{tool_config=}"
             )
+              
         formatted_tools = None
-        if tools:
+        code_execution_tool = GoogleTool(code_execution=CodeExecution())
+        if tools == [code_execution_tool]:
```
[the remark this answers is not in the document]
asimchoudhary: Agree. The current code treats `code_execution` as an exclusive tool, preventing its use with others like `grounding`. I’ll raise a PR to update formatted_tools so it can detect and format multiple tool types together. The approach will follow the existing invoke() pattern, modifying it to expose parameters for each tool (e.g., 
```python
invoke(code_execution=True, grounding=True)
``` 
 ensuring code_execution can coexist with other Google tools in the same request. Let me know your thoughts on this approach!

PR #789, opened 2025-03-17
genai: Add image output parsing support to ChatGoogleGenerativeAI

windkit:
## PR Description

Add image output parsing support to ChatGoogleGenerativeAI when using Gemini 2.0 Flash Experimental with image generation. Also updated the README.md to provide an example.

## Type

🆕 New Feature

## Testing(optional)

 - Integration Test

## Note(optional)

 - Handle of generation errors such as stopped by safety is a TODO.

### DIFF
--- a/libs/genai/tests/unit_tests/test_function_utils.py
+++ b/libs/genai/tests/unit_tests/test_function_utils.py
@@ -105,9 +105,9 @@ def test_tool_with_array_anyof_nullable_param() -> None:
     assert isinstance(items_property, dict), "Expected a dict."
 
     # Assertions
-    assert items_property.get("type_") == glm.Type.ARRAY, (
-        "Expected 'items' to be ARRAY."
-    )
+    assert (
+        items_property.get("type_") == glm.Type.ARRAY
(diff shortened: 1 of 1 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -455,6 +455,27 @@ def _parse_response_candidate(
             elif text:
                 raise Exception("Unexpected content type")
 
+        if part.inline_data.mime_type.startswith("image/"):
+            import base64
+
+            base64_data = base64.b64encode(part.inline_data.data).decode("utf-8")
```
lkuligin, line 461: can we use `_image_utils.image_bytes_to_b64_string` instead, please?

lkuligin (changes requested):

libs/genai/langchain_google_genai/chat_models.py
windkit: sure! update.

lkuligin (approved):

windkit:
@lkuligin updated with lint, PTAL.

PR #793, opened 2025-03-19
vertex[patch]: support mistral-small-2503 in model garden

ccurme:


lkuligin:
enabled in the test project

ccurme:
@lkuligin I'm still seeing tests fail with
> not allowed to use Publisher Model `projects/llm-integration-tests/locations/us-central1/publishers/mistralai/models/mistral-small-2503@001`"

Can you confirm it's enabled for the correct project / region?

libs/vertexai/Makefile
```diff
@@ -6,7 +6,7 @@ all: help
 # Define a variable for the test file path.
 TEST_FILE ?= tests/unit_tests/
 
-integration_test integration_tests: TEST_FILE = tests/integration_tests/
+integration_test integration_tests: TEST_FILE = tests/integration_tests/test_model_garden.py
```
lkuligin, line 9: is it by mistake?

Issue #764, opened 2025-03-03
ChatAnthropicVertex Claude 3.7 Thinking mode support


clouddev-code:
How ChatAnthropicVertex enable Claude 3.7 thinking mode?

The folowing Keyword arguments should  pass to the model.

`{"thinking": {"type": "enabled", "budget_tokens": 1024}}`

```
    llm = ChatAnthropicVertex(
        project="xxxxxxxxxxxxx",
        location=LOCATION,
        model_name="claude-3-7-sonnet@20250219",
        max_token=2049
    )
    try:
        chain =   prompt | llm | StrOutputParser()
        outputText = chain.invoke({"input":userMessage})
```

module version is the following
```
    "anthropic[vertex]>=0.49.0",
    "langchain-core>=0.3.40",
    "langchain-google-vertexai>=2.0.14",
```

DooGer1:
**How to Enable Claude 3.7 Thinking Mode in ChatAnthropicVertex**

To enable **Thinking Mode** in Claude 3.7 using **ChatAnthropicVertex**, you need to pass specific keyword arguments in the model's configuration. Below is the correct approach to achieving this.

### **Required Keyword Arguments**
The following dictionary should be passed to `model_kwargs` to enable Thinking Mode:

```json
{"thinking": {"type": "enabled", "budget_tokens": 1024}}
```

### **Updated Code Implementation**
Modify your ChatAnthropicVertex setup as follows:

```python
from langchain.chat_models import ChatAnthropicVertex
from langchain.schema import StrOutputParser

llm = ChatAnthropicVertex(
    project="xxxxxxxxxxxxx",
    location=LOCATION,
    model_name="claude-3-7-sonnet@20250219",
    max_token=2049,
    model_kwargs={  # Enable Thinking Mode
        "thinking": {
            "type": "enabled",
            "budget_tokens": 1024
        }
    }
)

try:
    chain = prompt | llm | StrOutputParser()
    outputText = chain.invoke({"input": userMessage})
    print(outputText)
except Exception as e:
    print(f"Error: {e}")
```

### **Explanation of Changes**
1. **Added `model_kwargs`**: This parameter now includes the `thinking` key to enable Thinking Mode.
2. **Defined `budget_tokens`**: Allocates **1024 tokens** for deep reasoning before generating a response.
3. **Ensures Compatibility**: The configuration works with **Claude 3.7** under Anthropic's Vertex AI integration.

### **Module Version Requirements**
[...]

PR #797, opened 2025-03-21
Add model_kwargs to ChatAnthropicVertex to enable thinking mode

hzeus:
Add a `model_kwargs` field to `ChatAnthropicVertex` that can hold `thinking` params.

I tried using the Claude 3.7 thinking mode via vertex and could not find a way to enable it. I then stumbled over the comment in https://github.com/langchain-ai/langchain-google/issues/764#issuecomment-2694399270 which suggested a `model_kwargs` param which is also what the [anthropic community model](https://github.com/langchain-ai/langchain/blob/d8145dda95d5a3d7c3e9feea1ff30eef25b14f00/libs/community/langchain_community/llms/anthropic.py#L67) does.

However I am not really sure if this is the desired or approach or if I am overseeing an existing solution here - I assume that https://github.com/langchain-ai/langchain-google/pull/790 already required a solution for this?

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/model_garden.py
+++ b/libs/vertexai/langchain_google_vertexai/model_garden.py
@@ -193,6 +193,7 @@ class ChatAnthropicVertex(_VertexAICommon, BaseChatModel):
     model_config = ConfigDict(
         populate_by_name=True,
     )
+    model_kwargs: dict[str, Any] = Field(default_factory=dict)
 
     # Needed so that mypy doesn't flag missing aliased init args.
     def __init__(self, **kwargs: Any) -> None:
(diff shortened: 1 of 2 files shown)


libs/vertexai/langchain_google_vertexai/model_garden.py
```diff
@@ -231,13 +232,14 @@ def validate_environment(self) -> Self:
 
     @property
     def _default_params(self):
-        return {
+        d = {
```
lkuligin, line 235: let's use a more readable name for the variable, please

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/model_garden.py
hzeus: @lkuligin I did, and also added a unit test

PR #811, opened 2025-03-25
genai[patch]: remove warning emitted during structured output

ccurme:
Remove warning associated with change in behavior from https://github.com/langchain-ai/langchain-google/pull/772.

It's undesirable to emit warnings for users following intended usage patterns as it can lead to confusion.

I don't feel strongly about when we remove the warning, in case we want to keep it around for N more weeks. But I'm also OK to remove at this point.


lkuligin:
let's maybe wait for 1-2 weeks, if that's OK

but approved anyway so it can be merged anytime

Issue #748, opened 2025-02-24
Propagate `model_version` to response metadata


ccurme:
When a model is specified with an alias (e.g., `"gemini-pro"`), it can be helpful to return the precise model that ran in `message.response_metadata`.

The model name should be accessible in the `response` at this point: https://github.com/langchain-ai/langchain-google/blob/375c6b2d3c15d8897017c0b72c8f63af3541ffc9/libs/vertexai/langchain_google_vertexai/chat_models.py#L1430

Ideally we'd do this for sync/async invoke and stream.

PR #813, opened 2025-03-26
vertexai[patch]: return model name in response metadata when streaming

ccurme:
https://github.com/langchain-ai/langchain-google/issues/748


libs/vertexai/tests/integration_tests/test_evaluation.py
```diff
@@ -57,6 +57,7 @@ async def test_aevaluate_strings() -> None:
 
 
 @pytest.mark.extended
+@pytest.xfail(reason="TODO: investigate (started failing 2025-03-25).")
```
ccurme, line 60: This test started failing, appears unrelated as it is also failing in https://github.com/langchain-ai/langchain-google/pull/815.

PR #817, opened 2025-03-26
genai: Fix embed_query method to use the non-batch endpoint

stgoa:
## PR Description

Fix the embed_query method so it uses the non batch endpoint:

`post https://generativelanguage.googleapis.com/v1beta/{model=models/*}:batchEmbedContents`

Instead of the batch endpoint 

`post https://generativelanguage.googleapis.com/v1beta/{model=models/*}:embedContent`

## Type

🐛 Bug Fix

## Changes(optional)

embed_query method of GoogleGenerativeAIEmbeddings

## Testing(optional)

Updated the test "test_embed_query"


lkuligin:
why do you think it is a problem?
afaik, we were sending a batch with one element only, but it was working, wasn't it? or am I missing anything?

stgoa:
hi @lkuligin the different endpoints have different quotas and rate limits. For instance you cannot sent over 150 RPM to the batch endpoint whereas the limit for the non-batch one is 1,500. There might be a pricing difference and performance difference also as in the openai api, but I haven’t measured that

lkuligin:
ok, it makes sense, thx!

lkuligin (approved):

Issue #791, opened 2025-03-18
While using the "ChatGoogleGenerativeAI" class getting error object GenerateContentResponse can't be used in 'await' expression


chetanrudrakar:
While using the "ChatGoogleGenerativeAI" class getting error object GenerateContentResponse can't be used in 'await' expression
Issue Description :
While utilizing the ChatGoogleGenerativeAI class, I encountered a TypeError indicating that the object GenerateContentResponse cannot be used within an await expression. This error arises when attempting to invoke asynchronous methods such as ainvoke or astream on the ChatGoogleGenerativeAI instance.

Steps to Reproduce:

Initialize the ChatGoogleGenerativeAI class with the appropriate parameters, including the model and API key and transport set to rest.
Observe the TypeError indicating that GenerateContentResponse cannot be used in an await expression.
ERROR    [agent] ❌ Result failed 1/3 times:
 object GenerateContentResponse can't be used in 'await' expression
Note : I am using this class with the browser-use Agent class to perform a task. Can anyone help me with this, Please refer the below code for reference.

async def script_validation():
task = (
‘Important : I am UI Automation tester validating the tasks’
‘Open website https://www.google.co.in/’
)
gemini_api_key = os.environ[“GEMINI_API_KEY”]
llm = ChatGoogleGenerativeAI(model=‘gemini-2.0-flash’, api_key=SecretStr(gemini_api_key), transport=“rest”)
agent = Agent(task=task, llm=llm, use_vision=True)
result = await agent.run()

asyncio.run(script_validation())

Can anyone please help ?

​

​

asimchoudhary:
hii @chetanrudrakar , i have tried the ainvoke on ChatGoogleGenerativeAI instance with the await keyword and it is working fine , the issue might be related to the Browse use api , can you please give some more context.

chetanrudrakar:
@asimchoudhary Hello Asim,  
I’m using `browser-use` to interact with the browser through an LLM to perform tasks. However, I’m encountering an error and am unsure of the cause. Let me know if you need more details to understand the issue..

asimchoudhary:
@chetanrudrakar , can you please share the error  and the code which is causing the error

chetanrudrakar:
@asimchoudhary The error shown is "object GenerateContentResponse can't be used in 'await' expression" and the code is already mentioned in the post you just need to update a gemini key.

lkuligin:
you can't use `rest` transport with async endpoints at this point in time

[...]

PR #822, opened 2025-03-27
genai: added transport check for async client in ChatGoogleGenerativeAI class

imperorrp:
## PR Description

Added a transport check for GenAI like in VertexAI. Now if an async method like `ainvoke()` is called, the async client is created with a transport field automatically set to "grpc_asyncio" if "rest" was passed originally. 

The equivalent VertexAI implementation that this PR replicates for GenAI: https://github.com/langchain-ai/langchain-google/blob/480a8b277b08e9060537b0435fe7bebcc7dc413a/libs/vertexai/langchain_google_vertexai/_base.py#L171 

## Relevant issues

For #791 

## Type

🐛 Bug Fix

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -962,12 +962,17 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
         # this check ensures that async client is only initialized
         # within an asyncio event loop to avoid the error
         if not self.async_client_running and _is_event_loop_running():
+            # async clients don't support "rest" transport
+            # https://github.com/googleapis/gapic-generator-python/issues/1962
+            transport = self.transport
+            if transport == "rest":
+                transport = "grpc_asyncio"
             self.async_client_running = genaix.build_generative_async_service(
                 credentials=self.credentials,
                 api_key=google_api_key,
                 client_info=get_client_info("ChatGoogleGenerativeAI"),
                 client_options=self.client_options,
-                transport=self.transport,
+                transport=transport,
             )
         return self.async_client_running
 


lkuligin:
@imperorrp I'm sorry, but have you tested that it's working?

I might be missing something, but afaik if you're using a sync client, you can't use it with `await` expressions, so you need to add additional workaround everywhere we're invoking the client itself.

imperorrp:
> @imperorrp I'm sorry, but have you tested that it's working?
> 
> I might be missing something, but afaik if you're using a sync client, you can't use it with `await` expressions, so you need to add additional workaround everywhere we're invoking the client itself.

Yeah- if `ainvoke()` is called with transport set to `rest`, the "TypeError: object GenerateContentResponse can't be used in 'await' expression" error is avoided (due to this auto conversion of the transport). That's what the VertexAI implementation seems to do (linked above) and test for as far as I could find too- https://github.com/langchain-ai/langchain-google/blob/main/libs/vertexai/tests/integration_tests/test_chat_models.py#L796C1-L801C74 

I think I might have gotten confused about intended functionality here though- did you mean sync clients should be used even if a user uses an async method like `ainvoke`?

imperorrp:
Tested with this to check, works: 
```from langchain_google_genai import ChatGoogleGenerativeAI
import google.generativeai as genai
import os 
import asyncio

async def test_func():
    llm = ChatGoogleGenerativeAI(model="gemini-1.5-pro", transport="rest")
    #print(llm.invoke("Sing a ballad of LangChain."))
    result = await llm.ainvoke("Sing a ballad of LangChain.")
    print(result)

if __name__ == "__main__":
    asyncio.run(test_func())
```

Print statements here to show modified/unmodified transport value:
image

Output (before this change, with transport "rest" passed without changing):
image
-error log trimmed-
image

 Output (after this change, with the transport switched to "grpc_asyncio"):
image

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -964,12 +964,17 @@ def async_client(self) -> v1betaGenerativeServiceAsyncClient:
         # this check ensures that async client is only initialized
         # within an asyncio event loop to avoid the error
         if not self.async_client_running and _is_event_loop_running():
+            # async clients don't support "rest" transport
+            # https://github.com/googleapis/gapic-generator-python/issues/1962
+            transport = self.transport
+            if transport == "rest":
+                transport = "grpc_asyncio"
```
lkuligin, line 971: ok, I see now. It might be a problem who're using a http proxy that doesn't allow grpc.

I was thinking we should just use sync client in that case, but that's not how we're doing things currently. we can fix it later then if smbd complains.

lkuligin (approved):

PR #831, opened 2025-03-31
test(genai): add test for `ChatGoogleGenerativeAI` class transport check

imperorrp:
## PR Description

Added a test for PR #822 for parity with the VertexAI transport check test at https://github.com/imperorrp/langchain-google/blob/main/libs/vertexai/tests/integration_tests/test_chat_models.py#L788-L803 

## Relevant issues

#791 

## Type

✅ Test

## Testing

image


libs/genai/tests/integration_tests/test_chat_models.py
```diff
@@ -574,3 +574,22 @@ async def model_astream(context: str) -> List[BaseMessageChunk]:
...
+
+@pytest.mark.extended
+def test_prediction_client_transport():
+    model = ChatGoogleGenerativeAI(model=_MODEL)
+    assert model.client.transport.kind == "grpc"
+
+    model = ChatGoogleGenerativeAI(model=_MODEL, transport="rest")
+    assert model.client.transport.kind == "rest"
+
+    async def check_async_client():
```
lkuligin, line 587: why should this be async function?

imperorrp: `model.async_client` does not appear to exist outside of an async function

imperorrp: So if this is added before the async function:

```
model = ChatGoogleGenerativeAI(model=_MODEL)
print(model.async_client)
print(model.client)
assert model.async_client.transport.kind == "grpc_asyncio"
```

then the test results look like this:
image

mdrxy:
@imperorrp sorry for delay but can you please resolve conflicts? Otherwise if this is solved can we close?

Issue #659, opened 2024-12-19
What does "Key '$defs' is not supported in schema, ignoring" mean?


pinkponk:
I keep getting `Key '$defs' is not supported in schema, ignoring` when I use gemini flash 1.5 with structured output but I can't seem to understand what the warning means. Cannot find anything on the web. Anyone knows? I think it might have something to do with nested pydantic models

I use the `langchain-google-vertexai` pkg at version 2.0.9

RobSisson:
I've hit the same issue when doing the Langgraph Academy course in module 5 - I believe it occurs for me when using a nested schema in Trustcall with gemini-1.5-flash.

Not sure if this is gemini / vertex ai related, as I've hit some other bugs throughout the course when running the provided notebooks without changing anything other than the model.

One which I'll highlight is another one which I've not been able to solve and also relates to using Trustcall.

Error: 'Key 'examples' is not supported in schema, ignoring'

Circumstance: Using gemini-1.5-flash to update an existing collection using Trustcall

`# Invoke the extractor with our updated conversation and existing memories
result = trustcall_extractor.invoke({"messages": updated_conversation, 
                                                        "existing": existing_memories})`

Schema:
`from pydantic import BaseModel, Field

class Memory(BaseModel):
    content: str = Field(description="The main content of the memory. For example: User expressed interest in learning about French.")

class MemoryCollection(BaseModel):
    memories: list[Memory] = Field(description="A list of memories about the user.")`

Unsolved: I think it is relating to gemini struggling to respond when the with_structured_output function, and it continuously generates the "example" field value which trustcall can't deal with. 

I've tried changing the trustcall prompt and putting the temperature to 0, both with no luck.

RobSisson:
As a note - been playing around and all of the errors I faced working with Gemini were solved completely when using OpenAI instead.

pinkponk:
Yes running the openai models works much better with the structured output. I however would like to know what the warning actually means when using the gemini models

agercas:
The issue is indeed with the nested pydantic model. Google/VertexAI doesn't support pydantic BaseModel directly and it is not even clear based on the documentation if they support complex nested types with structured output

https://ai.google.dev/gemini-api/docs/structured-output?lang=python

If you use pydantic BaseModel, this package converts it to a custom google schema using one of these functions
[...]

PR #843, opened 2025-04-04
vertexai: Fix resolving of references in ChatVertexAI.with_structured_output

SebastianThomas1:
## PR Description

When using a JSON schema with references, the method `langchain_google_vertexai.chat_models.ChatVertexAI.with_structured_output` logs a warning, stating that the key `'$defs’` is not supported. This holds if the JSON schema is passed as a dictionary or a typed dict type, or if it is passed as a Pydantic model with the (default) keyword argument `method = None`. (The case of a Pydantic model with the keyword argument `method="json_mode"` is already handled.)

## Relevant issues

Fixes [#659](https://github.com/langchain-ai/langchain-google/issues/659).

## Type

🐛 Bug Fix

## Changes(optional)

- The implementation of the method `langchain_google_vertexai.chat_models.ChatVertexAI.with_structured_output` is changed such that in the case `method="json_mode"`, the helper function `langchain_google_vertexai._utils.replace_defs_in_schema` is called in all cases for the type of the parameter `schema` (not only if it is a Pydantic model).
- ~The implementation of the helper function `langchain_google_vertexai.function_utils._dict_to_gapic_schema` is changed such that an item for the key `"$defs"` in the parameter `schema` is deleted if it exists after resolving references.~ The implementation of the helper function `langchain_google_vertexai.function_utils._format_json_schema_to_gapic` is fixed such that the key `"$defs"` in the parameter `schema` is silently ignored (instead of `"definitions“`, which is the Pydantic v1 variant). (Remark: This covers the case `method=None` in `langchain_google_vertexai.chat_models.ChatVertexAI.with_structured_output`.)
- The method `langchain_google_vertexai.chat_models.ChatVertexAI.with_structured_output` is changed such that a value error is raised if the parameter `schema` is neither a dictionary nor a Pydantic model.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -1150,12 +1150,12 @@ class ChatVertexAI(_VertexAICommon, BaseChatModel):
 
     logprobs: Union[bool, int] = False
     """Whether to return logprobs as part of AIMessage.response_metadata.
-    
-    If False, don't return logprobs. If True, return logprobs for top candidate. 
+
+    If False, don't return logprobs. If True, return logprobs for top candidate.
     If int, return logprobs for top ``logprobs`` candidates.
-    
+
     **NOTE**: As of 10.28.24 this is only supported for gemini-1.5-flash models.
-    
+
     .. versionadded: 2.0.6
     """
     labels: Optional[Dict[str, str]] = None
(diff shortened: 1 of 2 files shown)


muriloime:
I think you should add 

raw_schema["defs"] , as per https://github.com/langchain-ai/langchain-google/issues/659#issuecomment-2704415987

SebastianThomas1:
> I think you should add
> 
> raw_schema["defs"] , as per [#659 (comment)](https://github.com/langchain-ai/langchain-google/issues/659#issuecomment-2704415987)

Thank you, I appreciate your comment!

In the linked comment, `raw_schema["$defs"]` is replaced by `raw_schema["defs"]`, because the ref-template was chosen to use `"defs"` instead of `"$defs"`, which is the key if the JSON schema is constructed from a Pydantic model. In the end, `raw_schema["defs"]` is popped again.

I don't know why this implementation was chosen by the commenter, but it feels a bit hacky to me and I would think that it is not necessary.

The method `with_structured_output` also accepts if the JSON schema is passed directly, and I guess that it implicitly assumes that the default key `"$defs"` is used for references in this case. If we want to allow other keys (e.g., the [OpenAPI documentation](https://swagger.io/docs/specification/v3_0/using-ref/) uses `"components/schemas"`), we should either pass this key as a keyword argument and use it in the implementation to resolve the references, or somehow infer it from the actual references. But this would be a separate ticket, I guess, since it isn't anymore about the bug fix (resolving refs and removing defs since Gemini doesn't allow them), but an extension of the functionality of `with_structured_output` (or the fix of another bug, depending on the point of view ;) ).

Do I miss something?

SebastianThomas1:
@lkuligin Could you review this small PR or tag a potential reviewer, please? I don't know who is responsible and saw that you are the only person who is currently reviewing PRs for this repo.

lkuligin:
@SebastianThomas1 thanks for the contribution!

could you add a unit test to _dict_to_gapic_schema to check / demonstrate that dereferencing works, please?

SebastianThomas1:
Hi @lkuligin, thank you for your review!

I tried to add a test for `_dict_to_gapic_schema` that is similar to the tests `test_format_json_schema_to_gapic` (for `_format_json_schema_to_gapic`) and `test_format_json_schema_to_gapic_v1` (for `_format_json_schema_to_gapic_v1`). This is rather cumbersome since it is not easily possible to compare `gapic.Schema` instances (using `==` doesn't test what I expect it to test, the existing tests also only check some parts, but not the interesting ones that are related to the nesting). Since I don't have the time to write a really "good" test, I decided to not include a test for `_dict_to_gapic_schema` at all.

Nevertheless, your request was helpful as it led me to a better solution to the problem. Instead of changing the implementation for `_dict_to_gapic_schema`, I now changed the implementation of `_format_json_schema_to_gapic`. In fact, the latter contained a little copy/paste error, which led to unintended behavior: The case `key == "definitions"` was silently ignored, and `key == "$defs"` was ignored with the warning. However, `key == "definitions"` is the Pydantic v1 case (the line is the same as in `_format_json_schema_to_gapic_v1`), the Pydantic v2 analogon is just `key == "$defs"`. So I just had to replace/fix the literal in order to silence the warning.

As mentioned above, the helper functions `_format_json_schema_to_gapic` and `_format_json_schema_to_gapic_v1` already have tests, so I didn't add new tests for them. Also, I didn't strengthen the existing tests by checking that a warning is raised or not raised for certain keys, since I don't think that this should be part of the specification of the functions (and hence it also shouldn't be tested): An alternative solution would be to just always raise a warning (or even an exception) and to ensure that only schemas with allowed keys are passed to this helper functions.

lkuligin (approved):

PR #844, opened 2025-04-05
vertexai: fix embedding property override issue in VectorStore

ykoh42:
## PR Description

Refactored the `embeddings` property handling in `VectorStore` to ensure proper overriding from parent class. The previous implementation had a typo(`embbedings`) or mismatch that caused the property to not be overridden correctly, leading to unexpected behavior when accessing `embeddings`.  

## Type

🐛 Bug Fix
🧹 Refactoring

## Changes(optional)

Ensured that the `embeddings` property is correctly overridden and accessed via `@property`.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
+++ b/libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
@@ -7,6 +7,7 @@ from google.cloud.aiplatform.matching_engine.matching_engine_index_endpoint impo
     NumericNamespace,
 )
 from google.oauth2.service_account import Credentials
+from langchain_core._api.deprecation import deprecated
 from langchain_core.documents import Document
 from langchain_core.embeddings import Embeddings
 from langchain_core.vectorstores import VectorStore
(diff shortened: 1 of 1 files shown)


libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
```diff
@@ -30,22 +30,22 @@ def __init__(
         self,
         searcher: Searcher,
         document_storage: DocumentStorage,
-        embbedings: Optional[Embeddings] = None,
+        embeddings: Optional[Embeddings] = None,
```
lkuligin, line 33: that's a great catch, I wonder how we missed it before :).

but let's make this change backwards compatible, please. just add a new argument without removing an old one, throw a warning that the old one would be removed soon and pass it as
`self._embeddings = embeddings or embbedings  or self._get_default_embeddings()`

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
ykoh42: Thanks for the feedback!
I’ve made the change backwards compatible as suggested — kept the old argument, added a warning, and updated the assignment accordingly.
Let me know if there’s anything else to adjust!

lkuligin (approved):

PR #860, opened 2025-04-15
vertexai: fix handling of single string in add_texts

ykoh42:
## PR Description

Fix an issue where passing a single string to `add_texts` would split it into individual characters instead of treating it as a single document. This happens because Python strings are iterable, so `list(texts)` would convert a string like "hello" into `["h", "e", "l", "l", "o"]`. Added a check to automatically wrap a single string in a list.

## Type

🆕 New Feature
🐛 Bug Fix
✅ Test

## Changes(optional)

- Added type checking in `add_texts` to detect single strings and wrap them in a list
- Added unit test to verify the fix works as expected

## Testing(optional)

Added a new unit test `test_add_texts_with_single_string` that verifies a single string gets properly wrapped in a list before being processed.

## Note(optional)

This bug would cause quiet failures where users passing a single string would get unexpected results - the string would be split into individual characters and each character would be embedded as a separate document. The fix ensures a more intuitive API behavior where strings are handled as single documents.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
+++ b/libs/vertexai/langchain_google_vertexai/vectorstores/vectorstores.py
@@ -281,7 +281,11 @@ class _BaseVertexAIVectorStore(VectorStore):
 
         # Makes sure is a list and can get the length, should we support iterables?
         # metadata is a list so probably not?
-        texts = list(texts)
+        if isinstance(texts, str):
+            texts = [texts]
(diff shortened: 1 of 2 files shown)


lkuligin (approved):

lkuligin:
Since the signature is `texts: Iterable[str],`, I'm not sure a user should even pass a single string :)

lkuligin (approved):

PR #871, opened 2025-04-17
vertexai: loosen dependency on validators package

artdent:
## PR Description

Because the `validators` project is <1.0, the version restriction `^0.34` does not allow any other minor versions. This is unnecessarily restrictive and causes dependency conflicts with other packages. Because there have been no breaking changes to `validators.url`, the version restriction can be loosened to allow any recent version of the validators package.

## Relevant issues

n/a

## Type

🐛 Bug Fix


artdent:
Tests are failing on cloud build, but unfortunately the output on GitHub is truncated and I don't have permission to see the full output. Tests pass locally for me.

lkuligin (approved):

PR #884, opened 2025-04-23
genai: Added thinking_budget support to ChatGoogleGenerativeAI.

### LINKED ISSUE #872: Add thinking_budget parameter to ChatGoogleGenerativeAI to support Gemini 2.5 Flash
The new release of Gemini 2.5 Flash has introduced a new parameter to Google's SDK, as follows:

prompt = "Explain the Occam's Razor concept and provide everyday examples of it"
budget = 1024 # You can set this variable to any value between 0 and 24k

client = genai.Client(api_key="your-api-key")

response = client.models.generate_content(
  model="gemini-2.5-flash-preview-04-17",
  contents=prompt,
  config=types.GenerateContentConfig(
    thinking_config=types.ThinkingConfig(
      thinking_budget=budget
    )
  )
)

print(response.text)

As seen in https://ai.google.dev/gemini-api/docs/thinking#set-budget

windkit:
## PR Description
Add thinking_budget support to ChatGoogleGenerativeAI.

## Relevant issues

Fixes #872 

## Type

🆕 New Feature
✅ Test

## Testing(optional)

 - Integration Test

## Note(optional)

- Require updated https://github.com/googleapis/google-cloud-python/tree/main/packages/google-ai-generativelanguage/google/ai/generativelanguage_v1beta


aaramg:
+1

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -623,14 +623,24 @@ def _response_to_result(
...
+        thought_tokens = response.usage_metadata.thoughts_token_count
         cache_read_tokens = response.usage_metadata.cached_content_token_count
         if input_tokens + output_tokens + cache_read_tokens + total_tokens > 0:
-            lc_usage = UsageMetadata(
-                input_tokens=input_tokens - prev_input_tokens,
-                output_tokens=output_tokens - prev_output_tokens,
-                total_tokens=total_tokens - prev_total_tokens,
-                input_token_details={"cache_read": cache_read_tokens},
-            )
+            if thought_tokens > 0:
```
lkuligin, line 629: what is it's None?

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1222,6 +1233,8 @@ def _prepare_params(
             }.items()
             if v is not None
         }
+        if self.thinking_budget is not None:
+            gen_config["thinking_config"] = {"thinking_budget": self.thinking_budget}
```
lkuligin, line 1237: please, add it the same way as other parameters are added (two lines above)

libs/genai/tests/integration_tests/test_chat_models.py
```diff
@@ -174,6 +175,39 @@ def test_chat_google_genai_invoke_with_modalities() -> None:
...
+    result = llm.invoke(
+        "How many O's are in Google? Please tell me how you double checked the result",
+    )
+
+    assert isinstance(result, AIMessage)
+    assert isinstance(result.content, str)
+
+    _check_usage_metadata(result)
+
+    assert "output_token_details" not in result.usage_metadata
```
lkuligin, line 209: we don't need a separate test for that, add it to an existing integration test, please

windkit: thinking model by default performs auto reasoning, so this test explicitly disables the thinking by setting the budget to 0.

libs/genai/langchain_google_genai/chat_models.py
windkit: the value from the proto should be 0 if it is unset.

Wazbat:
Once this is merged, is there a rough idea of when it'll be available in the js version of the library? Does it require a seperate PR?

tutu-sol:
@Wazbat A patch to js side would be needed. Given the underlying proto change has been released, the patch there should also be straight forward.
I am not familiar with the js side, it will take me some time to ramp up the dev environment for that, I will see what I can do there.

windkit:
@Wazbat it requires a patch to js library.

lkuligin (approved):

PR #891, opened 2025-04-25
genai, vertexai[patch]: support model_kwargs

ccurme:
Catch params on init that are invalid according to Pydantic and store in model_kwargs.


libs/vertexai/tests/unit_tests/__snapshots__/test_standard.ambr
```diff
@@ -15,6 +15,10 @@
       'max_output_tokens': 100,
       'max_retries': 2,
       'model_family': '2',
+      'model_kwargs': dict({
+        'api_key': 'test',
```
ccurme, line 19: There's no `api_key` parameter in langchain-google-vertexai. The model is being initialized with this param as part of a test. Previously these params were discarded.

lkuligin (approved):

PR #908, opened 2025-05-08
genai: Fix the issue task_type passed into embed_query is ignored.

### LINKED ISSUE #902: Bug in from `langchain_google_genai import GoogleGenerativeAIEmbeddings.embed_query`
The `task_type` passed to `langchain_google_genai import GoogleGenerativeAIEmbeddings.embed_query` is ignored. Refer this [line](https://github.com/langchain-ai/langchain-google/blob/main/libs/genai/langchain_google_genai/embeddings.py#L256).

This bug leads to an un-intuitive in behavior for the two following calls:
```python


embeddings = GoogleGenerativeAIEmbeddings(model="models/gemini-embedding-exp-03-07", task_type="clustering")
emb = embeddings.embed_query("How does alphafold work?", output_dimensionality=768)

embeddings2 = GoogleGenerativeAIEmbeddings(model="models/gemini-embedding-exp-03-07")
emb2 = embeddings2.embed_query("How does alphafold work?",  task_type="clustering", output_dimensionality=768)

# Check for point-wise equality
emb == emb2
# >>> False
```

windkit:
## PR Description

Fix the issue task_type passed into embed_query is ignored.

## Relevant issues

Fixes #902 

## Type

🐛 Bug Fix
✅ Test

### DIFF
--- a/libs/genai/langchain_google_genai/embeddings.py
+++ b/libs/genai/langchain_google_genai/embeddings.py
@@ -253,7 +253,10 @@ class GoogleGenerativeAIEmbeddings(BaseModel, Embeddings):
         Returns:
             Embedding for the text.
         """
-        task_type = self.task_type or "RETRIEVAL_QUERY"
+        # Determine the task_type to use for this specific query
+        task_type_to_use = task_type if task_type is not None else self.task_type
+        if task_type_to_use is None:
+            task_type_to_use = "RETRIEVAL_QUERY"  # Default to RETRIEVAL_QUERY
         try:
             request: EmbedContentRequest = self._prepare_request(
                 text=text,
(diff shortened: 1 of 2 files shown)


libs/genai/langchain_google_genai/embeddings.py
```diff
@@ -253,7 +253,10 @@ def embed_query(
         Returns:
             Embedding for the text.
         """
-        task_type = self.task_type or "RETRIEVAL_QUERY"
+        # Determine the task_type to use for this specific query
```
lkuligin, line 256: nits: comment is not needed

you can also just do ` task_type if task_type else self.task_type`, no need for is not None

PR #918, opened 2025-05-12
genai:Make legacy function calling kwarg `function_call` optional

lolamathematician:
Set default to False as it is legacy
Included and updated tests

## PR Description

Adds option to not include (legacy) function calls in additional_kwargs in addition to tool calls. Other providers can raise an error if these are present in their calls.

This can come up if providers are mixed in a tool-calling loop.

We add a new optional flag to the ChatGoogleGenerativeAI model. Default to False for forward looking change.

Updated tests and added specific test for this change.

## Type

🐛 Bug Fix
✅ Test

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -47,10 +47,9 @@ from google.ai.generativelanguage_v1beta.types import (
     GenerationConfig,
     Part,
     SafetySetting,
-    ToolConfig,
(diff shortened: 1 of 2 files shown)


lkuligin:
@lolamathematician linter if failing, could you double-check, please?

plus, I don't think we should move test outside of "integration" or "unit" folders.

mdrxy:
Closing as stale, feel free to re-open with scoped down requested changes

PR #937, opened 2025-05-21
genai: Added include_thoughts support.

windkit:
## PR Description

genai: Added include_thoughts support.

## Type

🆕 New Feature
✅ Test

### DIFF
--- a/libs/genai/langchain_google_genai/_common.py
+++ b/libs/genai/langchain_google_genai/_common.py
@@ -84,6 +84,11 @@ Supported examples:
         default=None, description="Indicates the thinking budget in tokens."
     )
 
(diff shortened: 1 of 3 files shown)


lkuligin (approved):

yunhzou:
Although the thinking is passed to the model in the payload as parts, when recently testing on gemini 3, the model doesn't seem to perceive it. Not sure if this is a google side issue

I tested with 
First human input: 
in your reasoning say a very specific detail about a random fact but you should output an analysis on chemistry of caffeine. this is a test. 

Then: 
Tell me exactly the fact you mentioned in previous reasoning block.

The model is completely lost and have no idea about the content in previous thinking block.

PR #939, opened 2025-05-21
[community]: [remove arbitrary return in google results]

cuiqui:
## PR Description

This PR fixes the implementation to match the documented behavior of [this method](https://github.com/langchain-ai/langchain-google/blob/main/libs/community/langchain_google_community/search.py#L105).

Method's documentation clearly states that:

```
        Returns:
            A list of dictionaries with the following keys:
                snippet - The description of the result.
                title - The title of the result.
                link - The link to the result.
```

Much to my demise, there's a strange early return

```
        if len(results) == 0:
            return [{"Result": "No good Google Search Result was found"}]
```

if no results were found.

There's no need to represent the "emptiness" of results, an empty list takes care of that. In any case, if we need to represent the absence, this is not a sensible null pattern implementation, since the returned object is not even polymorphic wit h what the documentation says. 

This PR removes that early return and considers that an empty list correctly represents the absence of results

## Type

🧹 Refactoring
📖 Documentation

### DIFF
--- a/libs/community/langchain_google_community/search.py
+++ b/libs/community/langchain_google_community/search.py
@@ -125,8 +125,6 @@ class GoogleSearchAPIWrapper(BaseModel):
         results = self._google_search_results(
             query, num=num_results, **(search_params or {})
         )
-        if len(results) == 0:
-            return [{"Result": "No good Google Search Result was found"}]
         for result in results:
             metadata_result = {
                 "title": result["title"],


libs/community/langchain_google_community/search.py
```diff
@@ -125,8 +125,6 @@ def results(
         results = self._google_search_results(
             query, num=num_results, **(search_params or {})
         )
-        if len(results) == 0:
```
lkuligin, line 128: this is a breaking change.

please, add a corresponding note to the documentation, and add a deprecation warning first (that the behaviour would change after XX/XX/2025).

mdrxy:
Closing due to no movement, feel free to ping and I can re-open

PR #950, opened 2025-05-28
[vertexai] Fix image support for ToolMessage in langchain-vertexai

rong-xyz:
Basically, try send a base64 image (say from MCP) in ToolMessage you will easily get 900k token usage, which a lot of the user would assume it will just work if they transfermed from anthropic
This is a fix where the model behaves relatively good by faking role: function into a user but with tags
I made a similar fix to langchain-aws , where the model will simply skip translation if it is a ToolMessage, here everything is to_str() and image is easily 900k token usage https://github.com/langchain-ai/langchain-aws/pull/495 

As confimed by someone worked in Google:
> Right now this is somewhat undefined, but model will probably figure it out if you send the image as a user role, and document this in the dev instruction
a63e4a45282ce9ecadd09418bd848b0
end quote.

```
The google data structure is a pain to use and work with compared to other providers, It will be too much work to write a test, as I will use this to patch our code base anyways, I will submit the PR here 
```

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -399,19 +399,13 @@ def _parse_chat_history_gemini(
 
             vertex_messages.append(Content(role=role, parts=parts))
         elif isinstance(message, ToolMessage):
-            role = "function"
-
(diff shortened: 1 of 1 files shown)


mdrxy:
following up on this @rong-xyz

PR #982, opened 2025-06-16
community: added model armor sanitization runnables 

mihirvala08:
## PR Description

Community runnables to sanitize/screen prompt and LLM response in Langchain application using Google Model Armor. These runnables are derived from Lang Chain runnable interface to allow them integrating in any lang chain application flow for screening and sanitizing prompt and/or llm response.

More detail technical design can be referred here: [go/modelarmor-langchain-tdd](https://goto.google.com/modelarmor-langchain-tdd)

## Type

🆕 New Feature
✅ Test

## Changes

- Added Model Armor user prompt sanitization runnable
- Added Model Armor LLM response sanitization runnable
- Added Unit Tests for runnables
- Added Integration tests for runnables
- Added samples of different usage of runnables in Lang chain application


harshnasitcrest:
Could someone help in enabling Model Armor API and granting the `modelarmor.admin` role to the project that is used in running integration tests?

libs/community/langchain_google_community/model_armor/samples/callback_event_handler_usage.py
```diff
@@ -0,0 +1,96 @@
+"""
+Example usage of Model Armor runnables with LangChain's callback event handler
```
lkuligin: we do not put samples here, please, create example notebooks in LC documentation section

harshnasitcrest:
I've removed the samples from this PR. I'll create a new directory "model_armor" under https://github.com/langchain-ai/langchain/tree/master/docs/docs/integrations and add the notebooks.

@lkuligin Could you please confirm whether this is the intended location to submit PRs with Python notebooks as part of the documentation?

libs/community/langchain_google_community/model_armor/base_runnable.py
```diff
@@ -0,0 +1,167 @@
...
+    SanitizationResult,
+)
+from langchain_core.callbacks.manager import dispatch_custom_event
+from langchain_core.messages import BaseMessage
+from langchain_core.prompts import BasePromptTemplate
+from langchain_core.runnables import Runnable
+from langchain_core.runnables.config import RunnableConfig
+
+
+def setup_default_logging() -> logging.Logger:
```
lkuligin: why do we need this?

libs/community/langchain_google_community/model_armor/base_runnable.py
```diff
@@ -0,0 +1,167 @@
...
+        fail_open: If True, allows unsafe prompts/responses to pass through.
+        return_findings: If True, returns dict with prompt/response with
+                    the model armor sanitization findings.
+        logger: Logger instance for logging events. if not provided,
+            a default logger is created with INFO level.
+    """
+
+    def __init__(
+        self,
+        client: ModelArmorClient,
```
lkuligin: please, do not pass a client from the outside but create it during initiation of the class

libs/community/langchain_google_community/model_armor/runnable.py
```diff
@@ -0,0 +1,148 @@
...
+
+
+class ModelArmorSanitizePromptRunnable(ModelArmorSanitizeBaseRunnable):
+    """
+    Runnable to sanitize user prompts using Model Armor.
+    """
+
+    def invoke(
+        self,
+        input: Any,
```
lkuligin: should input be `LanguageModelInput`?

https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/language_models/base.py#L85

libs/community/langchain_google_community/model_armor/runnable.py
```diff
@@ -0,0 +1,148 @@
...
+            else:
+                raise ValueError("Prompt flagged as unsafe by Model Armor.")
+
+        if self.return_findings:
+            return {
+                "prompt": input,
+                "findings": sanitization_findings,
+            }
+
+        return input
```
lkuligin, line 116: please, make returned type consistent, not depending on init parameters?

it should return either a dict, or a PromptValue.

libs/community/langchain_google_community/model_armor/runnable.py
```diff
@@ -0,0 +1,148 @@
...
+class ModelArmorSanitizePromptRunnable(ModelArmorSanitizeBaseRunnable):
+    """
+    Runnable to sanitize user prompts using Model Armor.
+    """
+
+    def invoke(
+        self,
+        input: Any,
+        config: Optional[RunnableConfig] = None,
+        **kwargs: Any,
```
lkuligin, line 65: please, allow fail_open and return_findings args to be passed here as well (and override values that were provided during init)

lkuligin (changes requested):

libs/community/langchain_google_community/model_armor/README.md
```diff
@@ -0,0 +1,27 @@
+# Model Armor Runnables for LangChain
+
+This package provides LangChain-compatible Runnables for user prompt and model response sanitization using [Google Cloud Model Armor](https://cloud.google.com/security-command-center/docs/model-armor-overview).
+
+## Prerequisites
```
lkuligin: ideally, that should be a part of a docstring since then it would make it to the official documentation

I'm afraid not many people would find the Readme otherwise

libs/community/langchain_google_community/model_armor/samples/callback_event_handler_usage.py
harshnasitcrest: Noted. We've already removed samples from this PR in the last commit.

libs/community/langchain_google_community/model_armor/base_runnable.py
harshnasitcrest: I've replied to this on design doc.

There are multiple ways to create and init ModelArmorClient. Ref: https://cloud.google.com/python/docs/reference/google-cloud-modelarmor/latest/google.cloud.modelarmor_v1.services.model_armor.ModelArmorClient.html

We're currently using ADC for auth to Google APIs. Other users can use service account JSON file and use that to create the clients. Consequently, we have delegated this decision to the developers.

Furthermore, a single client object can be reused for all Model Armor runnables.

Let us know your thoughts.

libs/community/langchain_google_community/model_armor/base_runnable.py
harshnasitcrest: If no logger object is provided while initialising the Model Armor runnables, this method will be used to create a default logger object. 

Please refer to the [Logging](https://docs.google.com/document/d/1Z1ZoRaquifSxpLkuxeeWuqvVVpiSDh2phLoa8YVd-Ac/edit?tab=t.0#heading=h.8swio0g6ej28) section in TDD for more details.

libs/community/langchain_google_community/model_armor/runnable.py
harshnasitcrest: Type defined in abstract method of `Runnable` class is `Input` which is a `Generic` type. Ref: https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/runnables/utils.py#L41

Also, `_extract_input` function converts the input data to string (as required by Model Armor Sanitization API request). Ref: [SanitizeUserPromptRequest.user_prompt_data](https://github.com/googleapis/google-cloud-python/blob/848d82340281340679bb68979db799d2583a4a12/packages/google-cloud-modelarmor/google/cloud/modelarmor_v1/types/service.py#L1057), [DataItem.text](https://github.com/googleapis/google-cloud-python/blob/848d82340281340679bb68979db799d2583a4a12/packages/google-cloud-modelarmor/google/cloud/modelarmor_v1/types/service.py#L1512)

libs/community/langchain_google_community/model_armor/base_runnable.py
lkuligin: I'd like all integrations to have the same look and feel. Please, take a look at ChatVertexAI, you can pass an optional credential object (or path to service account json during initialization).

Initiating a client should take a few ms, so I don't see a need to re-use it (since the Runnable should be easily serializable). Anyways, if have a strong reason to pass a client, feel free to add a classmethod, please, but still support client creation during initialization (and also add a test that a runnable can be serialized and de-serialized in that case, please).

harshnasitcrest: Got it! Thanks for providing the reference, I'll take a look and update the PR accordingly.

libs/community/langchain_google_community/model_armor/base_runnable.py
harshnasitcrest: Removed and used logger by taking reference from https://github.com/langchain-ai/langchain-google/blob/567206e3f8e1da7fd6bd9bb502d438ab87971cdd/libs/vertexai/langchain_google_vertexai/llms.py#L20

libs/community/langchain_google_community/model_armor/runnable.py
harshnasitcrest: Done.
Removed return_findings flag as discussed with Utsav offline.

libs/community/langchain_google_community/model_armor/runnable.py
harshnasitcrest: Update return type. After screening, the input will be returned as is.

libs/community/langchain_google_community/model_armor/base_runnable.py
harshnasitcrest: Updated

libs/community/langchain_google_community/model_armor/README.md
harshnasitcrest: Done. Moved this to runnable.py file.

PR #985, opened 2025-06-18
Fix issue with thought_signature introduced in 1.96.0

czgu:
## PR Description

In https://github.com/googleapis/python-aiplatform/pull/5380, Google introduced a new field called thought_signature.
We need to exclude thought_signature similar to thought because v1Beta and v1 protobuf are defined differently.

In v1Beta, thought_signature field can co-exist with text field, but in v1 they are exclusive to each other.

## Relevant issues

None

## Type
🐛 Bug Fix

## Note(optional)
This issue will occur when google-cloud-aiplatform>=1.96.0 is installed.


czgu:
fixed in https://github.com/langchain-ai/langchain-google/pull/991, this PR is not needed

PR #1003, opened 2025-06-25
genai: fix multi-turn conversations with code execution blocks

### LINKED ISSUE #1002: ChatGoogleGenerativeAI fails with Multi-Turn conversations when using code execution
`ChatGoogleGenerativeAI` breaks in multi-turn conversations when the AI response includes code execution blocks. After the first turn, where the model successfully returns a response with executable code and its result, subsequent turns fail with an error: `Unrecognized message part type: executable_code. Only text, image_url, and media types are supported.`

This issue prevents maintaining a conversational context when using code execution tools, as the `AIMessage` containing code execution blocks cannot be passed back to the model for further interactions.

1. Run the following Python code:
```python
from google.ai.generativelanguage_v1beta.types import Tool as GenAITool

llm = ChatGoogleGenerativeAI(
    model="gemini-2.0-flash",
)

def chat_with_execution():
    messages = []

    while True:
        user_input = input("\nYou: ")
        if user_input.lower() == 'exit':
            break

        messages.append(HumanMessage(content=user_input))
        try:
            response = llm.invoke(messages, tools=[GenAITool(code_execution={})])
            print(f"\nAI: {response.content}")
            messages.append(response)
        except Exception as e:
            print(f"\nError occurred: {e}")

chat_with_execution()
```
2. Provide the following inputs: 
   - First input: `What's 8^8?`
   - Second input: `hi`
3. Observe the output:
```
You: What's 8^8?
AI: [{'type': 'executable_code', 'executable_code': 'print(8**8)\n', 'language': <Language.PYTHON: 1>}, {'type': 'code_execution_result', 'code_execution_result': '16777216\n'}, '8^8 is 16777216.']

You: hi
Error occurred: Unrecognized message part type: executable_code. Only text, image_url, and media types are supported.
```

#### Expected Behavior
The model should handle multi-turn conversations correctly, even when previous `AIMessage` responses contain code execution blocks. Subsequent inputs (e.g `hi`) should not raise an error, and the conversation should continue seamlessly, maintaining the context of prior messages.

#### Actual Behavior
After the first turn, where the model returns a response with code execution blocks, any subsequent turn fails with the error: `Unrecognized message part type: executable_code. Only text, image_url, and media types are supported.`

#### Additional Context
- The issue appears to stem from the `AIMessage` response containing structured content (`executable_code` and `code_execution_result`) that the integration does not properly serialize or handle in subsequent turns.
- This behavior is specific to multi-turn conversations; single-turn interactions with code execution tools work as expected.
- The error message suggests that message handling does not appropriately handle the `executable_code` part type when passed back as part of the conversation history.

#### Possible Workaround
Currently, no reliable workaround exists, as stripping the code execution blocks from the `AIMessage` before appending it to the message history would lose context and break the conversation flow. A potential fix might involve:
- Updating the integration to handle `executable_code` and `code_execution_result` parts correctly in multi-turn scenarios.

msaadg:
## PR Description

This PR fixes a bug in `ChatGoogleGenerativeAI` where multi-turn conversations fail when an `AIMessage` contains code execution blocks (e.g `executable_code` or `code_execution_result`). The error `Unrecognized message part type: executable_code` occurred because the `_convert_to_parts` function did not handle these part types. The change updates `_convert_to_parts` to support `executable_code` and `code_execution_result`, enabling seamless multi-turn interactions with code execution tools.

## Relevant issues

Fixes #1002 

## Type

🐛 Bug Fix

## Changes
- Modified `langchain_google_genai/chat_models.py` in the `_convert_to_parts` function:
  - Added handling for `executable_code` part type by mapping to `Part(executable_code=ExecutableCode(...))`.
  - Added handling for `code_execution_result` part type by mapping to `Part(code_execution_result=CodeExecutionResult(...))`.
  - Included validation for required keys (`executable_code`, `language`, `code_execution_result`) to prevent malformed inputs.
- Ensured backwards compatibility by preserving existing logic for other part types (`text`, `image_url`, `media`, etc.).

## Testing
**Test Procedure**:
  - Ran the original reproduction script provided in the issue with `gemini-2.0-flash`.
  - Input sequence:
    1. `What's 8^8?` (triggers code execution).
    2. `hi` (continues conversation).
  - Verified that the conversation continues without errors and maintains context.
- **Test Result**:
  - No `Unrecognized message part type` errors occurred.
  - The model correctly processes multi-turn conversations with code execution blocks.
- Ran `make format`, `make lint`, and `make test` from the `genai` package root to ensure no formatting or test failures were introduced.

## Note
- The change is minimal and focused on the `genai` package, with no impact on other packages (`vertexai`, `community`).
- No new dependencies were added.
- The fix is backwards compatible, as it only adds support for new part types without altering existing behavior.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -282,35 +282,6 @@ def _convert_to_parts(
             if _is_lc_content_block(part):
                 if part["type"] == "text":
                     parts.append(Part(text=part["text"]))
-                elif part["type"] == "tool_use":
-                    if part.get("text"):
-                        parts.append(Part(text=part["text"]))
-                elif part["type"] == "executable_code":
-                    if "executable_code" not in part or "language" not in part:
-                        raise ValueError(
-                            "Executable code part must have 'code' and 'language' keys,"
-                            f" got {part}"
-                        )
-                    parts.append(
-                        Part(
-                            executable_code=ExecutableCode(
-                                language=part["language"], code=part["executable_code"]
-                            )
-                        )
-                    )
-                elif part["type"] == "code_execution_result":
-                    if "code_execution_result" not in part:
-                        raise ValueError(
-                            "Code execution result part must have"
-                            f"'code_execution_result' key, got {part}"
-                        )
-                    parts.append(
-                        Part(
-                            code_execution_result=CodeExecutionResult(
-                                output=part["code_execution_result"]
-                            )
-                        )
-                    )
                 elif is_data_content_block(part):
                     if part["source_type"] == "url":
                         bytes_ = image_loader._bytes_from_url(part["url"])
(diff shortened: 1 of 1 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -280,6 +282,35 @@ def _convert_to_parts(
             if _is_lc_content_block(part):
                 if part["type"] == "text":
                     parts.append(Part(text=part["text"]))
+                elif part["type"] == "tool_use":
+                    if part.get("text"):
+                        parts.append(Part(text=part["text"]))
```
ccurme: What issue is this fixing?

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -280,6 +282,35 @@ def _convert_to_parts(
...
+                elif part["type"] == "code_execution_result":
+                    if "code_execution_result" not in part:
+                        raise ValueError(
+                            "Code execution result part must have"
+                            f"'code_execution_result' key, got {part}"
+                        )
+                    parts.append(
+                        Part(
+                            code_execution_result=CodeExecutionResult(
+                                output=part["code_execution_result"]
```
ccurme: I think we need `outcome` here as well.

ccurme:
Thank you!

PR #1012, opened 2025-06-30
genai: release 2.1.6

ccurme:


libs/genai/tests/integration_tests/test_standard.py
```diff
@@ -55,10 +55,6 @@ def test_usage_metadata_streaming(self, model: BaseChatModel) -> None:
     def test_bind_runnables_as_tools(self, model: BaseChatModel) -> None:
         super().test_bind_runnables_as_tools(model)
 
-    @pytest.mark.xfail(reason=("investigate"))
```
ccurme, line 58: These are passing now.

PR #1016, opened 2025-07-02
genai[patch]: fix totals in usage metadata and refactor

ccurme:
Addresses two issues:

1. `output_tokens` in [usage_metadata](https://python.langchain.com/api_reference/core/messages/langchain_core.messages.ai.UsageMetadata.html) currently does not include thought token counts. Here we add those so that `input_tokens` + `output_tokens` = `total_tokens`.

2. refactor to use utility functions for adding and subtracting usage metadata in core.

### DIFF
--- a/libs/genai/tests/integration_tests/test_standard.py
+++ b/libs/genai/tests/integration_tests/test_standard.py
@@ -21,8 +21,9 @@ class TestGeminiAI2Standard(ChatModelIntegrationTests):
     @property
     def chat_model_params(self) -> dict:
         return {
-            "model": "models/gemini-2.0-flash-001",
+            "model": "models/gemini-2.5-flash",
             "rate_limiter": rate_limiter,
+            "thinking_budget": 0,
         }
 
     @property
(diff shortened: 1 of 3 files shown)


libs/genai/tests/integration_tests/test_standard.py
```diff
@@ -45,16 +46,6 @@ def supports_pdf_inputs(self) -> bool:
     def supports_audio_inputs(self) -> bool:
         return True
 
-    @pytest.mark.xfail(
-        reason="Likely a bug in genai: prompt_token_count inconsistent in final chunk."
-    )
```
ccurme, line 50: This bug remains for Gemini 1.5 or 2.0, where the usage metadata (raw from the SDK) looks like:
```
candidates_token_count: 11
candidates_token_count: 11
candidates_token_count: 11
candidates_token_count: 11
candidates_token_count: 10
```
The final number is consistent with `total_tokens - output_tokens`, so we take that as ground truth.

PR #1025, opened 2025-07-09
vertexai[patch]: support thought signatures in function calls

ccurme:
Gemini docs for feature: https://ai.google.dev/gemini-api/docs/thinking#signatures

Here we just support for function calls. Resulting AIMessages keep an association between tool call IDs and thought signatures, stored as a base64 string:
```python
AIMessage(
    content=[...],
    tool_calls=[
        {
            "type": "tool_call",
            "id": "abc_123",
            "name": "my_tool",
            "args": {"param1": "value1", "param2": "value2"},
        }
    ],
    additional_kwargs={
        "__gemini_function_call_thought_signatures__": {
            "abc_123": _bytes_to_base64(b"thought signature bytes")
        }
    },
)
```
Thought signatures can also exist on text parts when (1) thinking is enabled, and (2) tools are bound. These will likely break interoperability of messages across chat models, so support there is blocked by an ongoing effort to move to standardized content blocks (high priority / ongoing now).

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -402,15 +402,16 @@ def _parse_chat_history_gemini(
                         if isinstance(thought_signature, str):
                             thought_signature = _base64_to_bytes(thought_signature)
                 function_call = FunctionCall({"name": tc["name"], "args": tc["args"]})
-                if thought_signature:
-                    parts.append(
-                        Part(
-                            function_call=function_call,
-                            thought_signature=thought_signature,
-                        )
+                parts.append(
+                    Part(
+                        function_call=function_call,
+                        **(
+                            {"thought_signature": thought_signature}
+                            if thought_signature
+                            else {}
+                        ),
                     )
-                else:
-                    parts.append(Part(function_call=function_call))
+                )
 
             if len(vertex_messages):
                 prev_content = vertex_messages[-1]
@@ -672,7 +673,7 @@ def _parse_response_candidate(
                         )
                     )
 
-            if hasattr(part, "thought_signature") and part.thought_signature:
+            if getattr(part, "thought_signature", None):
                 # store dict of {tool_call_id: thought_signature}
                 if isinstance(part.thought_signature, bytes):
                     thought_signature = _bytes_to_base64(part.thought_signature)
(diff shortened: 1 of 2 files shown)


libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -378,8 +391,26 @@ def _convert_to_parts(message: BaseMessage) -> List[Part]:
...
+                        _FUNCTION_CALL_THOUGHT_SIGNATURES_MAP_KEY, {}
+                    ):
+                        thought_signature = message.additional_kwargs[
+                            _FUNCTION_CALL_THOUGHT_SIGNATURES_MAP_KEY
+                        ][tool_call_id]
+                        if isinstance(thought_signature, str):
+                            thought_signature = _base64_to_bytes(thought_signature)
                 function_call = FunctionCall({"name": tc["name"], "args": tc["args"]})
-                parts.append(Part(function_call=function_call))
+                if thought_signature:
```
mdrxy: Could be written for conciseness:

```
parts.append(
    Part(function_call=function_call, **({"thought_signature": thought_signature} if thought_signature else {}))
)
```

libs/vertexai/langchain_google_vertexai/chat_models.py
```diff
@@ -635,10 +667,26 @@ def _parse_response_candidate(
                         invalid_tool_call(
                             name=function_call.get("name"),
                             args=function_call.get("arguments"),
-                            id=function_call.get("id", str(uuid.uuid4())),
+                            id=tool_call_id,
                             error=str(e),
                         )
                     )
+
+            if hasattr(part, "thought_signature") and part.thought_signature:
```
mdrxy: `if getattr(part, "thought_signature", None):` is more concise

libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -887,6 +887,34 @@ def test_chat_vertexai_gemini_thinking_auto_include_thoughts() -> None:
     _ = model.invoke([input_message, response, next_message])
 
 
+@pytest.mark.release
+def test_thought_signatures() -> None:
```
mdrxy, line 891: Maybe a docstring along the lines of:

"Verifies that thought signature byte blobs flow correctly through the entire Gemini to GAPIC to LangChain parsing and back into subsequent calls, without crashing or losing type safety"

lkuligin (approved):

PR #1027, opened 2025-07-10
Add test coverage for optional BigQuery dependency

pankaj-bind:
## PR Description
This update addresses the issue of limited test coverage for optional dependencies within the `langchain-google-community` package, using `BigQueryLoader` as the initial implementation.

Previously, modules with optional dependencies might not have been adequately tested for scenarios where those dependencies were missing, potentially leading to unexpected errors for users with minimal installations.

This change enhances the robustness of the package by:

1.  **Adding a Guarded Import:** The `BigQueryLoader` now checks for the `google-cloud-bigquery` package upon instantiation. If the dependency is not found, it raises a clear `ImportError` with instructions on how to install it.
2.  **Introducing a New Unit Test:** A dedicated unit test has been added to simulate an environment where `google-cloud-bigquery` is not installed. This test verifies that the `BigQueryLoader` raises the correct `ImportError`, ensuring this behavior is maintained and preventing future regressions.

This approach improves the reliability and user experience of the library, providing graceful failures and clear guidance when required dependencies are missing.

All existing and new tests pass successfully after these changes.

### DIFF
--- a/libs/community/langchain_google_community/bigquery.py
+++ b/libs/community/langchain_google_community/bigquery.py
@@ -1,14 +1,22 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, List, Optional
+from typing import TYPE_CHECKING, List, Optional, Any
 
 from langchain_core.document_loaders import BaseLoader
 from langchain_core.documents import Document
+from langchain_core.utils import guard_import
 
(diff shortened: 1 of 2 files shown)


pankaj-bind:
Hey @lkuligin, please review this PR

mdrxy (approved):

PR #1036, opened 2025-07-15
chore: added latest meta models

tiaan720:
## PR Description

Add support for two new Llama models to the `_LLAMA_MODELS` list in `model_garden_maas/_base.py`:
- "meta/llama-4-maverick-17b-128e-instruct-maas"
- "meta/llama-4-maverick-17b-16e-instruct-maas"

## Relevant issues

N/A

## Type

🆕 New Feature

## Changes

- Added the following model names to the `_LLAMA_MODELS` list:
  - "meta/llama-4-maverick-17b-128e-instruct-maas"
  - "meta/llama-4-maverick-17b-16e-instruct-maas"

## Testing

- Confirmed that the models are now recognized as valid Llama models in the codebase.

## Note

This PR only updates the list of supported Llama models. No other changes were


tiaan720:
@njoshi22, Thanks for catching that mistake. I have confirmed the llama-4-scout-17b-16e-instruct-maas works, so we should correct the mistake.

Issue #1032, opened 2025-07-14
Google Generative AI Vector Store outdated?


mdrxy:
https://python.langchain.com/api_reference/google_genai/google_vector_store.html

The Google GenAI Semantic Retriever API (a.k.a. Google Generative AI Vector Store in SDKs) has now been succeeded by Vertex AI Vector Search, a more robust, scalable, and fully managed vector database solution within Vertex AI.

PR #1077, opened 2025-08-06
genai: switch to google-genai sdk where possible

emhagman:
## Additional guidelines

- [X] PR title and description are appropriate
- [X] Necessary tests and documentation have been added
- [X] Lint and tests pass successfully
- [X] The following additional guidelines are adhered to:
  - Optional dependencies are imported within functions
  - No unnecessary dependencies added to pyproject.toml files (except those required for unit tests)
  - PR doesn't touch more than one package
  - ~Changes are backwards compatible~

## PR Description

Switches the client for `libs/genai` over to https://github.com/googleapis/python-genai. This is probably 95% done but I'd love a contributor's help to make sure I didn't miss anything. I am not sure of the best way to handle keeping both old and new clients in the code and supporting old/new features or if we should remove support for the old `embeddings.py` and switch to the new ones? I know ideally we don't have breaking changes but this seems large enough to warrant a `3.x` potentially?

The motivation behind this was to add `base_url` and support for `REST` in async environments so that this library can be used behind proxies (such as `litellm`, custom coded ones, etc) for tracking and rate-limiting purposes. However, in general, it seems like this is the library that will be supported by Google moving forward anyways.

All unit tests and integration tests passed for me locally.

## Relevant issues

None

## Type

🆕 New Feature

## Changes

Switched client over to https://github.com/googleapis/python-genai but had to keep the old client for `embeddings.py` since not everything is supported yet. That may not be an issue though if this is true https://github.com/langchain-ai/langchain-google/issues/1032

## Note

- The new client DOES NOT SUPPORT `grpc`, but Google (and the LLM world) seem to agree on REST for this
- The new client DOES NOT SUPPORT embeddings the same way so the old client is still in here for now
- This seems like a breaking change since I didn't convert the responses, etc into the GAPIC format (protobufs, etc) in case anyone relied on Google proprietary things in the `metadata` or `additional_kwargs` etc of `langchain` Pydantic models where things are just passed through directly

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -133,14 +133,17 @@ from langchain_google_genai._image_utils import (
 
 logger = logging.getLogger(__name__)
 
-_allowed_params_prediction_service = [
+_allowed_params_prediction_service_gapi = [
+    "request",
+    "timeout",
+    "metadata",
+    "labels",
+]
+
+_allowed_params_prediction_service_genai = [
     "model",
     "contents",
     "config",
-    # "request",
-    # "timeout",
-    # "metadata", # TODO: Add this back in
-    # "labels",
 ]
 
 _FunctionDeclarationType = Union[
@@ -216,6 +219,10 @@ def _chat_with_retry(generation_method: Callable, **kwargs: Any) -> Any:
         wait_exponential_max=kwargs.get("wait_exponential_max", 60.0),
     )
 
+    allowed_params = kwargs.get(
+        "allowed_params", _allowed_params_prediction_service_gapi
+    )
+
     @retry_decorator
     def _chat_with_retry(**kwargs: Any) -> Any:
         try:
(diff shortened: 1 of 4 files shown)


emhagman:
Google just announced the deprecation of the old libraries
https://ai.google.dev/gemini-api/docs/libraries#previous-sdks

November 30th, 2025

lkuligin:
thanks for the contribution!

we're not using the deprecated SDK, so we're good without this change.

if you could work on switching`VertexAIEmbeddings` to the new SDK that would be awesome, since afaik it's the only place where deprecation would hurt us

pzhao1799:
Thanks for all the hard work! My organization also has a model proxy and has been struggling to use langgraph with gemini for this exact issue. Would love to see this merged in soon - and wouldn't mind helping out if necessary. Wondering if getting the new client codepath flagged would be the easiest way.

lkuligin:
> Thanks for all the hard work! My organization also has a model proxy and has been struggling to use langgraph with gemini for this exact issue. Would love to see this merged in soon - and wouldn't mind helping out if necessary. Wondering if getting the new client codepath flagged would be the easiest way.

`ChatVertexAI` allows you to use a custom proxy, have you considered using it?

mdrxy:
@emhagman thank you a million for getting the ball rolling on this.

I'm unable to make contributions to your branch (due to org ownership), so I've picked this up in #1330 - will continue development there.

arindam-b:
@mdrxy @emhagman - If you use Vertex AI api key, it works same way as Google AI Studio.

PR #1085, opened 2025-08-18
Guard against unknown FinishReason values

dan-wit:
## PR Description

Add a defensive guard around `candidate.finish_reason` in the Vertex AI integration so that the client no longer crashes when the server returns an unknown enum value (e.g., 15).  
The change replaces the fragile `candidate.finish_reason.name` access with a safe lookup that returns `None` (or the raw integer as a string if desired) when the enum is unknown.

## Relevant issues

None (feature improvement / bug fix).

## Type

🐛 Bug Fix

## Changes(optional)

- Updated `langchain_google/_utils.py`:
  - Replaced `candidate.finish_reason.name` with a safe guard:
    ```python
    "finish_reason": (
        candidate.finish_reason.name
        if candidate.finish_reason and hasattr(candidate.finish_reason, "name")
        else None
    ),
    ```


sumeetattree:
Is there anything I can help with to move this forward? We have been running into this quite  frequently as Google Vertex is constantly hallucinating tool calls.

PR #1086, opened 2025-08-18
chore(vertex): update embedding models to use 'gemini-embedding-001' and 'imagen-3.0'

mdrxy:


libs/vertexai/langchain_google_vertexai/embeddings.py
```diff
@@ -93,7 +93,7 @@ def _missing_(cls, value: Any) -> "GoogleEmbeddingModelVersion":
         if "textembedding-gecko@003" in value.lower():
             return GoogleEmbeddingModelVersion.EMBEDDINGS_DEC_2023
         if (
-            "text-embedding-004" in value.lower()
+            "gemini-embedding-001" in value.lower()
```
nicolablago, line 96: look at line 105-106, with this change you are returning too early?

mdrxy: #1184

Issue #1115, opened 2025-09-05
Change `perform_literal_eval_on_string_raw_content` default to False


ccurme:
ChatVertexAI currently calls [ast.literal_eval](https://github.com/langchain-ai/langchain-google/blob/354b54aabf35f1fa670adc773e58aa1af559f7aa/libs/vertexai/langchain_google_vertexai/chat_models.py#L349) on all string inputs by default.

This behavior was implemented in support of https://github.com/langchain-ai/langchain-google/issues/399.

LangChain now supports [multimodal prompts](https://python.langchain.com/docs/how_to/multimodal_prompts/) so the feature is no longer needed.

The feature is problematic: if a user replies to a conversation with a float, we break:
```python
from langchain_google_vertexai import ChatVertexAI

llm = ChatVertexAI(model="gemini-2.0-flash")
llm.invoke("1.0")
```
> TypeError: 'float' object is not iterable

Although `literal_eval` won't execute code, I suspect there is still potential for abuse here.

We should change the default for [perform_literal_eval_on_string_raw_content](https://github.com/langchain-ai/langchain-google/blob/354b54aabf35f1fa670adc773e58aa1af559f7aa/libs/vertexai/langchain_google_vertexai/chat_models.py#L1472) to False.

PR #1122, opened 2025-09-11
vertexai: require opt-in for ast.literal_eval on input message strings

ccurme:
Resolves https://github.com/langchain-ai/langchain-google/issues/1115.

### DIFF
--- a/libs/vertexai/tests/integration_tests/test_chat_models.py
+++ b/libs/vertexai/tests/integration_tests/test_chat_models.py
@@ -266,13 +266,24 @@ def test_multimodal_media_inline_base64(file_uri, mime_type) -> None:
 
 @pytest.mark.release
 @pytest.mark.first
-@pytest.mark.xfail(reason="need a model supporting more than 1M input tokens")
 def test_multimodal_media_inline_base64_template() -> None:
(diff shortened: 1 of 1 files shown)


libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -282,15 +293,8 @@ def test_multimodal_media_inline_base64_template() -> None:
...
     media_base64 = base64.b64encode(blob.download_as_bytes()).decode()
-    media_message = {
-        "type": "media",
-        "data": media_base64,
-        "mime_type": mime_type,
-    }
-    text_message = {"type": "text", "text": "Describe the attached media in 5 words!"}
-    message = HumanMessage(content=[media_message, text_message])
     chain = prompt_template | llm
-    output = chain.invoke({"input": [message]})
```
ccurme, line 293: This doesn't look like it was done properly— we're formatting a message object into a prompt. So the LLM ends up seeing
```
"[HumanMessage(content=[{'type': 'media', 'data': 'SUQzBAAAAAABBlRYWFgAAA...
```
which is likely why the above xfail was added. Here I'm just demonstrating how the multi-modal prompt template can be done and removing the xfail.

PR #1137, opened 2025-09-12
fix(vertexai): add Union type support to V1 schema formatter

mdrxy:
Add proper `anyOf` handling to `_format_json_schema_to_gapic_v1` to ensure consistent formatting between Pydantic V1 and V2 Union types.

Base case is reached when a schema contains only `'type'`, `'title'`, and keys in `_ALLOWED_SCHEMA_FIELDS_SET`.


mdrxy:
This feels fine but would appreciate review to ensure it doesn't inadvertently break things downstream @lkuligin @ccurme

PR #1226, opened 2025-10-02
fix(community): Handle `UnicodeDecodeError` in GmailSearch and ensure header robustness

### LINKED ISSUE #1030: UnicodeDecodeError when decoding message body with non-UTF-8 encoding
Hi team,

First, thank you for your work on this integration — it's been very helpful.

I'm encountering a `UnicodeDecodeError` when using the `GmailSearch` tool to retrieve and parse messages. The error occurs in the `_parse_messages` method when the message body is decoded as UTF-8, but the actual encoding of the message is different (e.g., Latin-1 or Windows-1252). Here's the traceback:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xc1 in position 13503: invalid start byte
```

The issue seems to happen at this line (from `gmail/search.py`):
```
message_body = email_msg.get_payload(decode=True).decode("utf-8")
```
While multipart messages have a fallback to Latin-1 decoding, non-multipart messages are always decoded as UTF-8, which can cause the tool to crash when the content is in a different encoding.

To improve robustness, I recommend wrapping the decoding step with a fallback,  more general solution could try multiple encodings or use `errors="replace"` or `errors="ignore"` to avoid hard crashes on malformed characters.

Let me know if you'd like me to open a PR with a patch.

Thanks again!

RONAK-AI647:
## Description

This PR resolves a critical vulnerability in the **GmailSearch** tool's message parsing logic that caused the tool to crash when encountering real-world email data like encode-decode and fetching sender's `subject/from`.

## Relevant issues

Fixes #1030

## Type

🐛 Bug Fix
🧹 Refactoring

## Changes(optional)

1. **Fix 1 : `UnicodeDecodeError` (Fixes #1030)**

Added errors="replace" to the decoding call for non-multipart message bodies. This will gracefully replace the invalid bytes with a placeholder characters instead of raising a fatal` UnicodeDecodeError`

2. **Fix 2: Header ` NoneType`  Crash (This is a new thing i found while I was testing this issue)**

The code fetched headers `(email_msg["Subject"])` which returned None when the header was missing. A downstream function (either **clean_email_body** or later string processing) then attempted to call a string method (like .replace()) on that None object, causing an **AttributeError: 'NoneType'** object has no attribute 'replace'.

**solution:** Added a check to safely convert any missing header value (None) to an empty string ("") upon retrieval.

## Testing(optional)

To be added after maintainers' initial review .
## Note(optional)

1. PR only touches code within the community package.
4. Changes are backwards compatible.

### DIFF
--- a/libs/community/langchain_google_community/gmail/search.py
+++ b/libs/community/langchain_google_community/gmail/search.py
@@ -89,8 +89,8 @@ class GmailSearch(GmailBaseTool):
 
             email_msg = email.message_from_bytes(raw_message)
 
-            subject = email_msg["Subject"]
-            sender = email_msg["From"]
+            subject = email_msg["Subject"] or ""
+            sender = email_msg["From"] or ""
 
             message_body = ""
             if email_msg.is_multipart():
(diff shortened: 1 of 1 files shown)


libs/community/langchain_google_community/gmail/search.py
```diff
@@ -89,8 +89,8 @@ def _parse_messages(self, messages: List[Dict[str, Any]]) -> List[Dict[str, Any]
 
             email_msg = email.message_from_bytes(raw_message)
 
-            subject = email_msg["Subject"]
-            sender = email_msg["From"]
+            subject = email_msg["Subject"] or ""
```
lkuligin: nits: you can just use `email_msg.get("Subject", "")` here

RONAK-AI647: i agree , this is robust and more pythonic, i will update it

PR #1238, opened 2025-10-10
fix(vertex): `ToolMessage` is translated to `function` role which does not exist .

### LINKED ISSUE #1059: ToolMessage is translated to `function` role which does not exist
When a LangChain `ToolMessage` is translated to a `Content`, it is assigned the role `function` in [chatmodels.py](https://github.com/langchain-ai/langchain-google/blob/main/libs/vertexai/langchain_google_vertexai/chat_models.py). This role does not exist, only `user` and `model` are valid values. This applies to both [v1](https://cloud.google.com/vertex-ai/docs/reference/rpc/google.cloud.aiplatform.v1#content) and [v1beta1](https://cloud.google.com/vertex-ai/docs/reference/rpc/google.cloud.aiplatform.v1beta1#content).

Interestingly, this appears to throw no API error by itself, but if the `Content` sent to the API is malformed (e. g. by adding an invalid field to it before sending it to Gemini), the API returns

```
WARNING  - langchain_core.language_models.llms - Retrying langchain_google_vertexai.chat_models._acompletion_with_retry.<locals>._completion_with_retry_inner in 4.0 seconds as it raised BadRequest: 400 POST https://.../models/gemini-2.5-flash:generateContent: Please use a valid role: user, model..
```

Either way, setting the role to `function` results in undefined behavior. [Google's docs](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#invoke-api) suggest to use the `user` role instead.

RONAK-AI647:
## Description

This PR resolves issue #1059 where the LangChain adapter for Google Gemini incorrectly translates the `ToolMessage` (and the legacy `FunctionMessage`) into a Gemini `Content` object using the role `“function”`.
Whereas "according to gemini docs, "function" is not a valid role , only `user` and `model` are accepted valid roles.

## Relevant issues

Fixes #1059 

## Type

🐛 Bug Fix
🧹 Refactoring

## Changes(optional)

1.  **Role Correction:** Changed the assigned role for both `FunctionMessage` and `ToolMessage` from the invalid `“function”` to the API-compliant **`“user”`** and `**model**`
2.  **Logic Correction:** Updated the conditional logic in the message merging section to check for `prev_content.role == "user"`. This ensures that consecutive tool responses are correctly merged into a single `“user”` turn, maintaining the strict alternating sequence required by the Gemini chat models.

## Note(optional)
This ensures tool-calling flows are correctly translated and accepted by the Gemini API, especially when the agent is passing tool results back to the model for the final time.
This will not throw the

 ``` WARNING  - langchain_core.language_models.llms - Retrying langchain_google_vertexai.chat_models._acompletion_with_retry.<locals>._completion_with_retry_inner in 4.0 seconds as it raised BadRequest: 400 POST https://.../models/gemini-2.5-flash:generateContent: Please use a valid role: user, model..``` now.

### DIFF
--- a/libs/vertexai/tests/unit_tests/test_chat_models.py
+++ b/libs/vertexai/tests/unit_tests/test_chat_models.py
@@ -414,7 +414,7 @@ def test_parse_history_gemini_function() -> None:
         name=tool_call_2["name"], args=tool_call_2["args"]
     )
 
-    assert history[2].role == "model"
+    assert history[2].role == "user"
     assert history[2].parts[0].function_response == FunctionResponse(
         name=fn_name_1,
         response={"content": message3.content},


RONAK-AI647:
@lkuligin @mdrxy , take a look.

lkuligin (approved):

PR #1248, opened 2025-10-16
feat(vertex): support standard content

ccurme:
Requires https://github.com/langchain-ai/langchain/pull/33521

### DIFF
--- a/libs/vertexai/tests/integration_tests/test_chat_models.py
+++ b/libs/vertexai/tests/integration_tests/test_chat_models.py
@@ -1506,7 +1506,7 @@ def test_multimodal_pdf_input_gcs(multimodal_pdf_chain: RunnableSerializable) ->
 @pytest.mark.release
 def test_multimodal_pdf_input_url(multimodal_pdf_chain: RunnableSerializable) -> None:
     # TODO: parallelize with gcs and b64 tests
-    url = "https://abc.xyz/assets/95/eb/9cef90184e09bac553796896c633/2023q4-alphabet-earnings-release.pdf"
+    url = "https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf"
     # URL
     response = multimodal_pdf_chain.invoke({"image": url})
     assert isinstance(response, AIMessage)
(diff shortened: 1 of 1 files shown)


libs/vertexai/langchain_google_vertexai/_compat.py
```diff
@@ -0,0 +1,82 @@
+"""Go from v1 content blocks to VertexAI format."""
```
ccurme, line 1: Adapted from langchain_google_genai._compat

libs/vertexai/langchain_google_vertexai/_compat.py
```diff
@@ -0,0 +1,82 @@
...
+    Returns:
+        List of dictionaries in VertexAI content format.
+    """
+    new_content: list = []
+    for block in content:
+        block_dict = dict(block)  # (For typing)
+
+        # TextContentBlock
+        if block_dict["type"] == "text":
+            new_block = {"type": "text", "text": block_dict.get("text", "")}
```
ccurme, line 26: We translate to langchain v0 format here instead of native Vertex (v0 -> native is already taken care of inside ChatVertexAI)

libs/vertexai/langchain_google_vertexai/_compat.py
```diff
@@ -0,0 +1,82 @@
...
+                if "signature" in extras:
+                    new_block = {
+                        "thought": True,
+                        "text": block_dict.get("reasoning", ""),
+                        "thought_signature": extras["signature"],
+                    }
+                    new_content.append(new_block)
+                # else: skip reasoning blocks without signatures
+                # TODO: log a warning?
+            # else: skip reasoning blocks without extras
```
ccurme, line 45: @mdrxy I think there are two issues with how reasoning is done here.

First is that we should in general translate `reasoning` blocks to `thinking` blocks and pass them back. See the `test_chat_vertexai_gemini_thinking_auto_include_thoughts` test for an example of how to engage thinking output.

The second issue is that from what I can tell, thought signatures appear on two types of Gemini content
1. function calls
2. text blocks, including those without `thought: True`.

Examples:
```
contents {
  role: "model"
  parts {
    text: "Let's use the get_weather tool..."
    thought: true
  }
  parts {
    function_call {
      name: "get_weather"
      args {
        fields {
          key: "location"
          value {
            string_value: "London"
          }
        }
      }
    }
    thought_signature: "..."
  }
}
```
---
```
content {
    role: "model"
    parts {
      text: "I want to convey helpfulness and that I'm ready to assist further when needed..."
      thought: true
    }
    parts {
      text: "You're welcome! If you have any other questions, feel free to ask."
      thought_signature: "..."
    }
  }
```
So my first instinct would be to include these as `extras` on tool_call blocks and text blocks. Right now we search for them on `thinking` blocks, but from what I can tell this is never engaged: https://github.com/langchain-ai/langchain/blob/31718492c7aaa9430691696808a66f5053bf7a39/libs/core/langchain_core/messages/block_translators/google_genai.py#L424-L431

PR #1274, opened 2025-11-01
fix(genai): Set `propertyOrdering` in schema by default when performing structured output

### LINKED ISSUE #1105: Set propertyOrdering in schema by default when performing structured output
Do we know if the current implementation of langchain-google uses the [propertyOrdering](https://ai.google.dev/gemini-api/docs/structured-output#property-ordering) schema from google gemini API?

Can we implement the [default ordering to generate ](https://github.com/langchain-ai/langchain-google/blob/063f9ee06789e1291ffef8ba6452290e8dedd694/libs/genai/langchain_google_genai/chat_models.py#L1903)the value of this field to be the order of fields in the pydantic object? 

This is how google genai client [internally does it](https://github.com/googleapis/python-genai/blob/214d01790b5fa9c008e150f0ffb77c4c363d9124/google/genai/_transformers.py#L781) as well.

This would improve quality of generations as well when properties have a useful ordering and its natural to define the order within the pydantic object rather than to define it explicitly with this property. Langchain should handle this internally

RONAK-AI647:
## Description

Langchain-google did not have the implementatation of propertyOrdering Schema from google gemini API.
This PR implements the automatic generation of the` propertyOrdering` field when using Pydantic models with Gemini's JSON Mode **(json_mode or json_schema).**

It introduces a new` boolean` parameter,` force_property_ordering`, to the **ChatGoogleGenerativeAI.with_structured_output method.**
## Relevant issues

Fixes #1105
## Type

🆕 New Feature
🧹 Refactoring

## Changes(optional)

1) New Parameter added ( force_property_ordering: bool = False)
2) Made Pydantic Field Extraction: Inside the json_mode/json_schema
3) : The extracted field_order list is inserted into the generated schema_json dictionary using the correct API key: schema_json['propertyOrdering'] = field_order.

### DIFF
--- a/libs/community/langchain_google_community/bq_storage_vectorstores/_base.py
+++ b/libs/community/langchain_google_community/bq_storage_vectorstores/_base.py
@@ -89,17 +89,19 @@ class BaseBigQueryVectorStore(VectorStore, BaseModel, ABC):
         filter: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
     ) -> List[Document]:
-        """Search documents by their ids or metadata values.
+        """Search documents by their IDs or metadata values.
 
         Args:
-            ids: List of ids of documents to retrieve from the vectorstore.
+            ids: List of IDs of documents to retrieve from the `VectorStore`.
             filter: Filter on metadata properties, e.g.
-                            {
-                                "str_property": "foo",
-                                "int_property": 123
-                            }
+                ```json
+                {
+                    "str_property": "foo",
+                    "int_property": 123
+                }
+                ```
         Returns:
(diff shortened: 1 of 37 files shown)


RONAK-AI647:
@lkuligin , @mdrxy , take a look

mdrxy:
I think this may no longer be necessary with the new improvements just announced to structured output via their API -- see #1288

Maybe useful for old version?

RONAK-AI647:
@mdrxy , I see the latest changes. Yeah , it could work for the older version but the new changes are best in itself. You can go ahead and close this pr .

PR #1288, opened 2025-11-05
feat(genai): improved structured output

mdrxy:
https://blog.google/technology/developers/gemini-api-structured-outputs/
https://ai.google.dev/gemini-api/docs/structured-output


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -2430,7 +2548,7 @@ def with_structured_output(
         self,
         schema: Union[Dict, Type[BaseModel]],
         method: Optional[
-            Literal["function_calling", "json_mode", "json_schema"]
+            Literal["function_calling", "json_mode", "json_schema", "json_schema_v2"]
```
ccurme: I'd consider controlling this with another kwarg (or even just replacing json_schema with v2 if we're not losing any functionality). we effectively standardize around these methods; `"json_schema"` typically means "provider-native feature". so we could just update the implementation to use v2, or introduce a separate kwarg that would let you toggle between v1 and v2.

libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -1810,8 +1892,26 @@ class Joke(BaseModel):
...
+    response_json_schema: Optional[Dict[str, Any]] = None
+    """Enhanced structured output schema supporting unions (`anyOf`) and recursive
+    schemas (`$ref`).
+    
+    !!! warning
+    
+        When both `response_schema` and `response_json_schema` are provided,
+        `response_json_schema` takes precedence.
+    
+    Replaces `response_schema` for expanded JSON Schema support including:
```
ccurme: here too, we could just update the implementation in a minor release. I'm less opposed to introducing another attribute here, since these are already specific to Google GenAI, but wondering if it's necessary to maintain both.

PR #1293, opened 2025-11-07
fix: `base_url` behavior, and improve checks and docstrings for `base_url` and `api_transport` behavior

mdrxy:


libs/vertexai/langchain_google_vertexai/_base.py
```diff
@@ -179,6 +180,7 @@ def async_prediction_client(
                 endpoint_version=self.endpoint_version,
                 credentials=self.credentials,
                 client_options=cast("ClientOptions", self.client_options),
+                transport=self.api_transport or "grpc",
```
jacoblee93: Do we need a default value here?

libs/vertexai/langchain_google_vertexai/_client_utils.py
```diff
@@ -64,16 +64,27 @@ def _create_async_prediction_client(
...
+    # https://github.com/googleapis/gapic-generator-python/issues/1962
+    # However, when using custom endpoints, we can try to keep REST transport
+    has_custom_endpoint = bool(
+        client_options.api_endpoint
+        and client_options.api_endpoint != "https://aiplatform.googleapis.com"
+    )
+
+    # Only change to grpc_asyncio if no custom endpoint is specified
+    if transport == "rest" and not has_custom_endpoint:
+        transport = "grpc_asyncio"
```
jacoblee93, line 96: Is the default value `grpc_asyncio` if you wanted to opt into gRPC?

PR #1298, opened 2025-11-09
fix(genai): Add validation for empty contents in Gemini API requests

yashwantbezawada:
## Summary

Fixes issue where invoking `ChatGoogleGenerativeAI` with only a `SystemMessage` causes a cryptic 400 error from the Gemini API.

**Related Issue:** langchain-ai/langchain-google#1351

## Problem

The Gemini API requires at least one non-system message (content) in the request. When only a `SystemMessage` is provided:
1. The message gets extracted as `system_instruction` in `_parse_chat_history()`
2. The `contents` array becomes empty
3. The Gemini API rejects the request with: `"Invalid argument provided to Gemini: 400 * GenerateContentRequest.contents: contents is not specified"`

## Solution

This PR adds validation after message parsing to check if the `history` (contents) array is empty and raises a clear `ValueError` with actionable guidance:

```python
# Validate that we have at least one content message for the API
if not history:
    msg = (
        "No content messages found. The Gemini API requires at least one "
        "non-system message (HumanMessage, AIMessage, etc.). "
        "If you only have a SystemMessage, please set "
        "convert_system_message_to_human=True when initializing the model, "
        "or include additional messages in your input."
    )
    raise ValueError(msg)
```

## Changes

- Added validation in `_prepare_request()` method (lines 2353-2362)
- Added two unit tests:
  - `test_system_message_only_raises_error()` - Verifies the error is raised
  - `test_system_message_with_convert_flag_works()` - Verifies the workaround works

## Testing

The fix includes comprehensive unit tests that verify:
1. A `ValueError` is raised when only a `SystemMessage` is provided
2. The error message guides users to the correct solution
3. The workaround (`convert_system_message_to_human=True`) works correctly

## Backward Compatibility

This change is backward compatible:
- No changes to existing successful use cases
- Only affects the edge case of passing only a `SystemMessage`, which previously failed with a cryptic error
- Now fails fast with a helpful error message

### DIFF
--- a/libs/genai/tests/integration_tests/test_chat_models.py
+++ b/libs/genai/tests/integration_tests/test_chat_models.py
@@ -27,7 +27,7 @@ from langchain_google_genai import (
     Modality,
 )
 
-_MODEL = "gemini-flash-lite-latest"
+_MODEL = "models/gemini-2.5-flash"
 _VISION_MODEL = "models/gemini-2.0-flash-001"
 _IMAGE_OUTPUT_MODEL = "models/gemini-2.0-flash-exp-image-generation"
 _AUDIO_OUTPUT_MODEL = "models/gemini-2.5-flash-preview-tts"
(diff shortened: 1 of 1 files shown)


yashwantbezawada:
## Important Correction

I've updated the PR to fix an error in the original implementation:

**The `convert_system_message_to_human` flag does NOT help with single `SystemMessage` scenarios.**

After careful code review, I discovered that `convert_system_message_to_human` only merges the system instruction with a subsequent `HumanMessage` at position `i==1` in the message processing loop (see line 703). It does not convert a standalone `SystemMessage` into a `HumanMessage`.

### Changes Made:
1. **Updated error message** to remove the misleading suggestion about `convert_system_message_to_human`
2. **Updated test** from `test_system_message_with_convert_flag_works()` to `test_system_message_with_additional_message_works()` to test the correct usage pattern (SystemMessage + HumanMessage)

### Correct Solution:
Users encountering this error should include at least one additional message alongside any `SystemMessage`:
```python
# ✓ This works
ChatGoogleGenerativeAI(model="gemini-2.0-flash-lite").invoke([
    SystemMessage(content="You are a helpful assistant"),
    HumanMessage(content="Hello")
])
```

The updated error message now correctly states:
> "No content messages found. The Gemini API requires at least one non-system message (HumanMessage, AIMessage, etc.) in addition to any SystemMessage. Please include additional messages in your input."

Issue #1380, opened 2025-11-24
ChatAnthropicVertex: BlockType Image not correct for LangChain 1.X


qvalentin:
### Package (Required)

- [ ] langchain-google-genai
- [x] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
see PR
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

With LangChain 1.X the format for content blocks did change: https://docs.langchain.com/oss/python/migrate/langchain-v1?ref=blog.langchain.com&ajs_aid=0774ee46-591d-412c-9f8a-5a5e3c7d21e3&ajs_uid=e4ce8b88-c947-4b00-a445-d0c3eb21f3be#what-changed

This is currently not supported by _format_message_anthropic for images.

PR #1381, opened 2025-11-24
fix(vertexai): `ChatAnthropicVertex` support images with `langchain 1.X`

qvalentin:
fixes https://github.com/langchain-ai/langchain-google/issues/1380

### DIFF
--- a/libs/community/langchain_google_community/__init__.py
+++ b/libs/community/langchain_google_community/__init__.py
@@ -15,6 +15,10 @@ from langchain_google_community.calendar.toolkit import (
     GetCalendarsInfo,
     GetCurrentDatetime,
 )
+from langchain_google_community.callbacks import (
+    AsyncBigQueryCallbackHandler,
+    BigQueryCallbackHandler,
+)
 from langchain_google_community.docai import DocAIParser, DocAIParsingResults
 from langchain_google_community.documentai_warehouse import DocumentAIWarehouseRetriever
 from langchain_google_community.drive import GoogleDriveLoader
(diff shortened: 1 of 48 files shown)


libs/vertexai/tests/unit_tests/test_anthropic_utils.py
```diff
@@ -785,25 +786,11 @@ def test_format_messages_anthropic_with_mixed_messages() -> None:
...
-                            "source_type": "url",
-                            "url": "data:image/png;base64,/9j/4AAQSk",
-                        },
-                        {
-                            "type": "image",
-                            "source_type": "base64",
-                            "mime_type": "image/png",
-                            "data": "/9j/4AAQSk",
-                        },
-                        {"type": "image", "source_type": "id", "id": "1"},
```
mdrxy, line 806: we must continue to maintain support for these input formats; we should not remove these tests

libs/vertexai/tests/unit_tests/test_anthropic_utils.py
```diff
@@ -828,14 +815,6 @@ def test_format_messages_anthropic_with_mixed_messages() -> None:
...
                             },
                         },
-                        {
-                            "type": "image",
-                            "source": {
-                                "type": "base64",
-                                "media_type": "image/png",
-                                "data": "/9j/4AAQSk",
-                            },
-                        },
```
mdrxy, line 838: same here

libs/vertexai/tests/unit_tests/test_anthropic_utils.py
qvalentin: LC 1.X had the breaking change, but sure, it does not hurt us to be backward compatible here.

HerrIvan:
any chance of releasing this? it is really affecting our end users.
can we in the community help further or does it only depend on someone langchain-ai approving it?

lkuligin (approved):

Issue #1384, opened 2025-11-26
ProviderStrategy fails with JSONDecodeError for Gemini on VertexAI


RohanDisa:
### Package (Required)

- [x] langchain-google-genai
- [ ] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
"""
Minimal reproducible example showing ProviderStrategy JSONDecodeError with Gemini.

This script demonstrates:
1. ToolStrategy works perfectly ✅
2. ProviderStrategy fails with JSONDecodeError ❌
3. Both use the same model, schema, and prompt

Run with:
    export GCP_PROJECT_ID="your-project-id"
    python reproduce_provider_strategy_bug.py

Requirements:
    pip install langchain langchain-google-vertexai
"""
import asyncio
import os
from langchain_google_vertexai import ChatVertexAI
from langchain.agents import create_agent
from langchain.agents.structured_output import ProviderStrategy, ToolStrategy

# Simple JSON schema for testing
SCHEMA = {
    "type": "object",
    "properties": {
        "sentiment": {
            "type": "string",
            "enum": ["positive", "negative", "neutral"],
            "description": "The sentiment of the text"
        },
        "confidence": {
            "type": "number",
            "minimum": 0,
            "maximum": 1,
            "description": "Confidence score"
        }
    },
    "required": ["sentiment", "confidence"],
    "additionalProperties": False
}

PROMPT = "Analyze this review: 'This product is absolutely amazing!'"


async def test_tool_strategy_works():
    """✅ This works - proves setup is correct"""
    print("\n" + "=" * 80)
    print("TEST 1: ToolStrategy with Gemini")
    print("=" * 80)

    model = ChatGoogleGenerativeAI(
        model="gemini-2.5-flash",
        temperature=0.0,
        credentials=None,  # Force API key usage instead of GCP credentials
    )

    agent = create_agent(
        model=model,
        tools=[],
        response_format=ToolStrategy(schema=SCHEMA, handle_errors=True),
    )

    try:
        result = await agent.ainvoke({
            "messages": [{"role": "user", "content": PROMPT}]
        })

        print("✅ SUCCESS - ToolStrategy works!")
[...]

PR #1385, opened 2025-11-26
fix(genai): add `ProviderStrategy` `response_format` support

RohanDisa:
Description: Add support for response_format from ProviderStrategy by extracting OpenAI-style JSON schema and mapping it into response_json_schema for Gemini models. Automatically sets the correct response_mime_type when structured output is used.
Fixes issue: #1384


mdrxy:
deferring until completion of migration to unified SDK

PR #1387, opened 2025-11-26
fix(genai): use tool_call_schema to exclude injected arguments

cyhhao:
## Problem

When using tools that have injected parameters like `ToolRuntime`, `InjectedState`, or `InjectedStore`, Google Genai's tool conversion fails with:

```
PydanticInvalidForJsonSchema: Cannot generate a JsonSchema for core_schema.CallableSchema
```

This happens because `_format_base_tool_to_function_declaration` uses `tool.args_schema` which includes all parameters, including injected ones that contain `Callable` types that Pydantic cannot serialize to JSON schema.

## Solution

Use `tool.tool_call_schema` instead of `tool.args_schema`. The `tool_call_schema` property (from langchain-core's `BaseTool`) already handles filtering out injected arguments, making it the correct schema to use when binding tools to models.

This is consistent with how `langchain-core`'s `convert_to_openai_function` handles tools - it uses `tool.tool_call_schema` (see [`_format_tool_to_openai_function`](https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/utils/function_calling.py#L297)).

## Testing

Tested locally with a tool using `ToolRuntime` parameter - the error is resolved and the tool works correctly with Google Gemini models.

## Related

- langchain-core's `BaseTool.tool_call_schema` property: filters out injected arguments
- OpenAI function conversion already uses `tool_call_schema`


mdrxy:
deferring until completion of migration to unified SDK

mdrxy:
Hi, `langchain-google-genai 4.0.0` is going out now! Would you mind verifying that this issue persists in the new release, and rebasing this PR against `main` if so?

mdrxy:
Thanks for the contribution!

This PR/issue has been inactive for a while, so we're going to close it for now to keep things tidy.

If this is still relevant, feel free to reopen it or start a new discussion with updated context. We're happy to take another look.

Thanks again! 🙏

PR #1418, opened 2025-12-08
chore(community): bump lint max ver to python 3.14

spartandingo:


spartandingo:
as per the comment in `.github/workflows/_lint.yml`, looks like max lint version should be the max supported python version. currently it's still at 3.12. bumping to 3.14

PR #1419, opened 2025-12-08
release(community): 3.0.2

spartandingo:


spartandingo:
putting a release version so we can use python 3.14 with the library. not sure if that's cool @mdrxy or if you have other stuff you want in the release. thank you :-)

PR #1428, opened 2025-12-10
fix(genai): Add support for video URL format in HumanMessage content

josiahcoad:
The docs show using {"type": "video", "url": "...", "mime_type": "video/mp4"} format, but the _convert_to_parts function was missing a handler for this.

The handler converts video URLs to FileData(file_uri=...) format that Google's API expects. This must be placed BEFORE the is_data_content_block check to prevent video dicts from being caught by the image handler.

Fixes issue where video URLs were being downloaded as HTML bytes instead of being passed as file_uri to the API.

Tested and verified working with gemini-2.5-flash  and 3.0-pro model.

### DIFF
--- a/PR_INSTRUCTIONS.md
+++ b/PR_INSTRUCTIONS.md
@@ -1,0 +1,77 @@
+# Pull Request Instructions
+
+## Changes Made
+Added support for `{"type": "video", "url": "...", "mime_type": "video/mp4"}` format in `HumanMessage` content.
+
+## Branch
+`fix/video-url-support`
+
+## Next Steps
+
+### 1. Fork the repository (if you haven't already)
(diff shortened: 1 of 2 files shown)


josiahcoad:
Thanks for taking @mdrxy Cheers!

mdrxy:
Can you please re-open from a personal fork? Maintaners can't edit PRs made originating from other organizations

PR_INSTRUCTIONS.md
```diff
@@ -0,0 +1,77 @@
+# Pull Request Instructions
```
mdrxy, line 1: this should be deleted

mdrxy:
could you also resolve CI failures?

PR #1443, opened 2025-12-16
fix(genai): fix instance `safety_settings` not applied in `ChatGoogleGenerativeAI`

Konsti-s:
## Description

The primary fix now adds ChatModel instance safety_settings to the model invocation params.
As a side effect type hints of safety_settings have been modified/fixed.

## Type

🐛 Bug Fix

## Changes

- Include instance safety_settings during model invocation in langchain_google_genai/chat_models.py
- Modify _format_safety_settings in langchain_google_genai/chat_models.py
- Fix type in langchain_google_genai/_common.py

## Testing

Added a new test that fails without the fix.

## Warning

ChatGoogleGenerativeAI now no longer accepts google.genai.types.SafetySetting, but only dict[HarmCategory, HarmBlockThreshold] as *probably* was the intended usage. This actually introduces a breaking change, which I believe will go unnoted, as I assume *almost* everybody is using the intended format anyway.

### DIFF
--- a/libs/genai/tests/unit_tests/test_chat_models.py
+++ b/libs/genai/tests/unit_tests/test_chat_models.py
@@ -4724,3 +4724,148 @@ def test_image_config_no_instance_config_with_override() -> None:
 
     assert config.image_config is not None
     assert config.image_config.aspect_ratio == "16:9"
+
+
+# ============================================================================
+# OpenAI-style response_format and strict kwarg handling
+# Compatibility with langchain.agents.create_agent
+# ============================================================================
+
+
+def test_openai_style_response_format_json_object() -> None:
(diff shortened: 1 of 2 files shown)


libs/genai/tests/unit_tests/test_chat_models.py
```diff
@@ -168,6 +168,46 @@ def test_safety_settings_initialization() -> None:
     assert llm.model == f"{MODEL_NAME}"
 
 
+def test_safety_settings_passed_to_api_from_instance() -> None:
+    """Test that instance safety_settings are passed forward."""
```
mdrxy:
```suggestion
    """Test that instance `safety_settings` are passed forward."""
```

PR #1471, opened 2025-12-21
fix(genai): ensure model generation params pass through to the request

### LINKED ISSUE #1454: [ChatGoogleGenerativeAI] Model attributes like `max_output_tokens` and `temperature` not respected
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python

model = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash-lite",  # Also tried 2.0, and not-lite
    max_output_tokens=20,           # Also tried max_tokens=20
)
# model.max_output_tokens = 20      # Also tried this

# Generate a response with the new settings
response = model.invoke(
    "Write a social media post about how much you're learning about transformers.",
    # max_output_tokens=2,          # This works as expected
)

print("Finish reason: ", response.response_metadata["finish_reason"])
print("Output tokens: ", response.usage_metadata["output_tokens"])
```

```shell
No error, but this is the output of my code above:

Finish reason:  MAX_TOKENS
Output tokens:  20
```

### Description

1. Instantiate a model with `max_output_tokens` set, which is described as an attribute [here](https://reference.langchain.com/python/integrations/langchain_google_genai/ChatGoogleGenerativeAI/#langchain_google_genai.ChatGoogleGenerativeAI.max_output_tokens).
2. Expect the response to be limited to the number of tokens.
3. The response is not limited to the number of tokens.

Setting `max_output_tokens` in the `.invoke()` does work as expected.

Running
```
langchain==1.2.0
google-genai==1.56.0
langchain-google-genai==4.1.1
```

I think this got introduced during the big overhaul in 4.0.0.

I experimented a bit and I think it happens in the `._prepare_request()` method or in the `._build_request_config`: most of the `params` are not carried over in the final request config. 

A quick fix that did the trick, was adding `max_output_tokens=self.max_output_tokens` in the config [here](https://github.com/langchain-ai/langchain-google/blob/2036835fea03e542d6ae21abaa08630904ef6354/libs/genai/langchain_google_genai/chat_models.py#L2804) or [here](https://github.com/langchain-ai/langchain-google/blob/2036835fea03e542d6ae21abaa08630904ef6354/libs/genai/langchain_google_genai/chat_models.py#L3004). But that breaks if `max_output_tokens` is set in the `invoke()` method. Also, this only handles this specific param, and I think there are more involved.

I'm happy to look further into it myself, but I probably won't have time in the coming days.

julesvanrie:
## Description

Since the move to the unified Google GenAI SDK in 4.0.0, model generation parameters like `max_output_tokens`, `temperature`, `top_k`, ... were not passed through to the actual request. They were created as part of the `params` object, but only select params were actually passed through.

This is an important fix, because currently the only way to pass these generation parameters is by adding them when calling the `invoke` method. However, this happens silently, and someone setting for example `temperature` at the model level, would not detect this immediately and hence would not be aware of the problem. Only `max_output_tokens` is relatively easy to detect when there are too many tokens in the output.

This PR fixes the issue, and makes the code easier to maintain, by making sure all `params` are passed through without having to enumerate them manually. This required moving some of the manipulation steps into the `_build_base_generation_config`  method where they sit more naturally.

## Relevant issues

Fixes #1454 

## Type

🐛 Bug Fix

## Changes

Pass through all `params` to `_build_request_config`.

To avoid having to enumerate all params in multiple places, leverage the model's Pydantic model dumping capabilities.

This required a couple of adjustments:
- When creating the `GenerateContentConfig` in `_build_request_config`, remove all parameters that are included in `params`. This allows us to consume the whole `params` object without duplicating parameters.
- This required moving some preparation steps for these parameters from this method to `_build_base_generation_config`, where they sit more naturally:
  - Convert `response_modalities` to their values.
  - Auto-set audio output for TTS models if not explicitly configured.
  - Create the `ThinkingConfig` object if needed.
- Change `_build_thinking_config` to return a `ThinkingConfig`` object
  instead of a dict, reducing the need to do the conversion later.

## Testing(optional)

- Changed the model initialization test to capture the failure, checking the config coming out of `_prepare_request` in line with other tests.
- Adapted the tests that check that model parameters can be overridden to also check `_prepare_request` rather than the earlier `_prepare_params` step. This to capture any regressions that might be introduced by changing the flow as described in the changes.

### DIFF
--- a/libs/genai/tests/integration_tests/test_chat_models.py
+++ b/libs/genai/tests/integration_tests/test_chat_models.py
@@ -2601,6 +2601,7 @@ def test_context_caching_tools(backend_config: dict) -> None:
     assert response.tool_calls[0]["name"] == "get_secret_number"
 
 
+@pytest.mark.flaky(retries=3, delay=1)
 def test_audio_timestamp(backend_config: dict) -> None:
     """Test audio transcription with timestamps.
 
@@ -2658,10 +2659,12 @@ def test_audio_timestamp(backend_config: dict) -> None:
     # - [00:00:00] or [HH:MM:SS] (with hours)
     # - [0:00-0:54] or [M:SS-M:SS] (time ranges)
     # - [ 0m0s405ms - 0m3s245ms ] (millisecond precision with brackets and spaces)
+    # - 0:00 - 7:12 (bare timestamp ranges without brackets)
     # Pattern matches various timestamp formats
     timestamp_patterns = [
         r"\[\d+:\d{2}",  # [0:00] or [00:00]
         r"\[\s*\d+m\d+s",  # [ 0m0s or [0m0s
+        r"\d+:\d{2}\s*-\s*\d+:\d{2}",  # 0:00 - 7:12 (time ranges)
     ]
     has_timestamps = (
         "**[" in text_content  # Format: **[00:00]**


julesvanrie:
@mdrxy Can you review this, despite one of the checks failing?

The last run of llm-integration-tests failed on `test_search_builtin`. It did not do this on a previous run, and running it locally did not result in errors either (tried it multiple times). I think re-running the check will probably result in passing the test.

The failure in the last run:

```txt
  File "/workspace/libs/genai/tests/integration_tests/test_chat_models.py", line 1563, in test_search_builtin
    _check_web_search_output(response, output_version)
  File "/workspace/libs/genai/tests/integration_tests/test_chat_models.py", line 1527, in _check_web_search_output
    assert len(text_blocks) >= 1
AssertionError: assert 0 >= 1
 +  where 0 = len([])
```

Local command used to test:

```bash
uv run pytest -k search_builtin -vv
```

Output:
```txt
tests/integration_tests/test_chat_models.py::test_search_builtin[google_ai-v0] PASSED [ 25%]
tests/integration_tests/test_chat_models.py::test_search_builtin[google_ai-v1] PASSED [ 50%]
tests/integration_tests/test_chat_models.py::test_search_builtin_with_citations[google_ai-False] PASSED [ 75%]
tests/integration_tests/test_chat_models.py::test_search_builtin_with_citations[google_ai-True] PASSED [100%]
```

PR #1472, opened 2025-12-22
feat(community): add BigQuery Callbacks

Genesis929:
## Description

This PR introduces BigQueryCallbackHandler and AsyncBigQueryCallbackHandler to log LangChain run lifecycle events (LLM, Chain, Tool, Agent, Retriever, and Chat Model activities) directly to Google Cloud BigQuery.

Key features include:

Logging of all standard LangChain event types (e.g., on_llm_start, on_tool_end, on_chain_error).

Automatic dataset and table creation during handler initialization if they don't exist.

Utilizes BigQuery's specialized write clients for both synchronous and asynchronous data ingestion.

## Type

🆕 New Feature
🐛 Bug Fix
🧹 Refactoring
📖 Documentation
🚄 Infrastructure
✅ Test

### DIFF
--- a/libs/community/langchain_google_community/callbacks/bigquery_callback.py
+++ b/libs/community/langchain_google_community/callbacks/bigquery_callback.py
@@ -49,6 +49,7 @@ logger = logging.getLogger(__name__)
 _GRPC_DEADLINE_EXCEEDED = 4
 _GRPC_INTERNAL = 13
 _GRPC_UNAVAILABLE = 14
+_DEFAULT_TRACE_ID = "langchain-bq-agent-analytics"
 
 
 def _recursive_smart_truncate(obj: Any, max_len: int) -> tuple[Any, bool]:
@@ -495,7 +496,7 @@ class _AsyncBatchProcessor:
             serialized_schema = self.arrow_schema.serialize().to_pybytes()
             serialized_batch = arrow_batch.serialize().to_pybytes()
             req = self.bq_storage_types.AppendRowsRequest(
-                write_stream=self.write_stream, trace_id="langchain-bq-callback"
+                write_stream=self.write_stream, trace_id=_DEFAULT_TRACE_ID
             )
             req.arrow_rows.writer_schema.serialized_schema = serialized_schema
             req.arrow_rows.rows.serialized_record_batch = serialized_batch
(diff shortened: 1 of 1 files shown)


haiyuan-eng-google:
@mdrxy  can you help take a look for this PR ? thanks

haiyuan-eng-google:
Run uv lock --check
Resolved 132 packages in 2.44s
The lockfile at `uv.lock` needs to be updated, but `--check` was provided. To update the lockfile, run `uv lock`.
Error: Process completed with exit code 1. @mdrxy is there something we can do to fix the CI/CD error here?  cc @Genesis929

Genesis929:
> * CI / cd libs/community / - / make test #3.12 (pull_request)

Seems this issue resolved, but currently we are seeing issue related to access, something like:

Building langchain-google-community @ file:///home/runner/work/langchain-google/langchain-google/libs/community
  × Failed to download `langsmith==0.4.42`
  ├─▶ Failed to fetch:
  │   `[https://us-python.pkg.dev/artifact-foundry-prod/ah-3p-staging-python/langsmith/langsmith-0.4.42-py3-none-any.whl`](https://us-python.pkg.dev/artifact-foundry-prod/ah-3p-staging-python/langsmith/langsmith-0.4.42-py3-none-any.whl%60)
  ╰─▶ HTTP status client error (401 Unauthorized) for url
      (https://us-python.pkg.dev/artifact-foundry-prod/ah-3p-staging-python/langsmith/langsmith-0.4.42-py3-none-any.whl)
  help: `langsmith` (v0.4.42) was included because
        `langchain-google-community` (v3.0.2) depends on `langchain-core`
        (v1.2.5) which depends on `langsmith`
        
  @mdrxy is there something we can do to fix this type of error?

mdrxy:
If `_log` encounters an exception during content parsing (e.g., in `parser.parse_message_content()`), the event appears to be silently dropped.

Also, the new callback handlers are importable from `langchain_google_community.callbacks` but not from the main package `__init__.py`. Should we add these?

Is there any room to easily de-duplicate some of the repeated logic? e.g. `_BatchProcessor._prepare_arrow_batch` and sync `_SyncBatchProcessor._prepare_arrow_batch` methods are nearly identical but the async version has more comprehensive JSON handling for string/bytes. This inconsistency could cause different behavior between sync and async handlers.

Appears to be missing error handling in `on_llm_end`.

I'd also recommend not using `assert` for runtime checks. Consider using explicit `if` checks.

Can we also standardize on using `X | None` instead of `Optional[]`?

The sync `shutdown()` sets `_is_shutting_down = False` at the end, which is inconsistent with the async version that uses a try/finally block. If an error occurs during shutdown, the flag may remain `True` in the async version but would be reset to `False` in the sync version.

haiyuan-eng-google:
fix the integration test

haiyuan-eng-google:
add a specific label for stream write api from bq callback via langchain, differentiate the trace between langchain and ADK  plugin

Genesis929:
> fix the integration test

Our test is skipped without additional env variable.

The error is from test_googlesearch_api.py and test_vertex_ai_search.py.

I assume it's not because of our change?

PR #1476, opened 2025-12-27
feat(genai): Standardize include_thoughts output format to fix LangSmith rendering

### LINKED ISSUE #1021: [VertexAI]: Text Output does not show up in Langsmith in combination with `include_thoughts`
When I use the newly introduced `include_thoughts` parameter, I can only see the thoughts part in Langsmith, but not the actual text output:

Image

I am not sure if this is an issue that should be fixed here or elsewhere. 

I looked into what was returned by calling `ainvoke` and noticed that `content` was a list containing a mix of `dict` and `str`:

```
[
  {'type': 'thinking', 'thinking': 'Okay, here\'s what\'s going through my head as I process this simple "Hi."\n\n(...)}, 
  'Hello! How can I help you today?'
]
```

I then tried to make the output more similar by updating https://github.com/langchain-ai/langchain-google/blob/main/libs/vertexai/langchain_google_vertexai/chat_models.py#L586-L594 to 

```python
elif text:
    text_message = {
        "type": "text",
        "text": text,
    }
    if not content:
        content = [text_message]
    elif isinstance(content, str):
        content = [text_message, content]
    elif isinstance(content, list):
        content.append(text_message)
    else:
        raise Exception("Unexpected content type")
```

This seems to have the desired effect - would it make sense to open a Pull Request for this?

RONAK-AI647:
## Description

This PR addresses an inconsistency in how `ChatGoogleGenerativeAI `handles response content when `include_thoughts=True.`
## Relevant issues

Fixes #1021 
## Type

🐛 Bug Fix
✅ Test

## Changes(optional)

`chat_models.py:` Modified logic to append text as a dictionary when thinking blocks are present.

`test_chat_models.py:` Updated legacy unit tests that were strictly asserting for raw strings. These tests now correctly assert for the standardized dictionary format.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -932,6 +932,12 @@ def _parse_response_candidate(
             if thought_sig or _is_gemini_3_or_later(effective_model_name or ""):
                 # append blocks if there's a signature or new Gemini model
                 content = _append_to_content(content, text_block)
+            elif isinstance(content, list) and any(
+                isinstance(item, dict) and item.get("type") == "thinking"
+                for item in content
+            ):
+                # if there's thinking blocks, keep content as dicts
+                content = _append_to_content(content, text_block)
             else:
                 # otherwise, append text
                 content = _append_to_content(content, text or "")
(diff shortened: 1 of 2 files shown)


RONAK-AI647:
@lkuligin @mdrxy take a review please !!

PR #1480, opened 2025-12-29
fix(vertexai): fix tests

lkuligin:


alx13 (approved):

libs/vertexai/tests/integration_tests/test_chat_models.py
```diff
@@ -849,7 +852,7 @@ def search(
     result = llm_with_search.invoke([request, response, *tool_messages])
 
     assert isinstance(result, AIMessage)
-    assert "brown" in result.content
+    assert "brown" in result.content[0]["text"]  # type: ignore
```
mdrxy, line 855: is this just for gemini 3+?

PR #1489, opened 2026-01-04
fix(genai): Prevent ADC lookup delay when initializing VertexAI with API key (optimized)

SyedImtiyaz-1:
## How the Fix Works

The fix prevents the ADC lookup delay by using `AnonymousCredentials` when an API key is provided but no credentials are present. 

**Implementation:**

1. **Credentials Detection**: Before initializing the Client, the code checks for:
   - `GOOGLE_APPLICATION_CREDENTIALS` environment variable
   - Default gcloud credentials file (OS-specific path)

2. **AnonymousCredentials Usage**: If no standard credentials exist and an API key is provided, `AnonymousCredentials()` is used to explicitly skip the ADC lookup that was causing the ~11 second delay.

3. **Project ID Optimization**: When no project ID is provided, a placeholder ("missing-project-id") is used to prevent slow project ID lookups.

4. **Backward Compatibility**: The fix only activates when no credentials are present. All existing authentication methods continue to work as before.

**Key Code Changes:**
- Added `AnonymousCredentials` import from `google.auth.credentials`
- Modified `validate_environment` method to detect credential presence
- Updated `Client()` initialization to use prepared credentials and project ID

This approach ensures the SDK doesn't perform unnecessary ADC lookups when API key authentication is being used, reducing initialization time from **~11s to <0.02s.**

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -21,6 +21,7 @@ from typing import (
 )
 
 import filetype  # type: ignore[import-untyped]
+from google.auth.credentials import AnonymousCredentials
 from google.genai.client import Client
 from google.genai.errors import ClientError
 from google.genai.types import (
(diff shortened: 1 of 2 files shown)


SyedImtiyaz-1:
Hi @mdrxy ,
Can you please review this one !

PR #1501, opened 2026-01-12
fix(genai): Normalize streaming token format for Gemini models

### LINKED ISSUE #1460: `on_llm_new_token` receives list instead of string with Gemini 3 models
### Checked other resources

- [X] I added a very descriptive title to this issue.
- [X] I searched the LangChain documentation with the integrated search.
- [X] I used the GitHub search to find a similar question and didn't find it.
- [X] I am sure that this is a bug in LangChain rather than my code.

### Example Code

```python
from langchain_core.callbacks import BaseCallbackHandler

class StreamingCallback(BaseCallbackHandler):
    def on_llm_new_token(self, token: str, **kwargs) -> None:
        print(f"Token type: {type(token)}, value: {token}")

llm = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash",  # or gemini-3-flash-preview
    streaming=True,
    callbacks=[StreamingCallback()]
)

response = llm.invoke("What is 2+2?")
```

### Error Message and Stack Trace

No error, but unexpected token format:

**With Gemini 2.5 Flash** (expected):
```
Token type: <class 'str'>, value: The
Token type: <class 'str'>, value:  answer
Token type: <class 'str'>, value:  is
Token type: <class 'str'>, value:  4
```

**With Gemini 3 Flash Preview** (unexpected):
```
Token type: <class 'list'>, value: [{'type': 'text', 'text': 'The', 'index': 0}]
Token type: <class 'list'>, value: [{'type': 'text', 'text': ' answer', 'index': 0}]
Token type: <class 'list'>, value: [{'type': 'text', 'text': ' is', 'index': 0}]
Token type: <class 'list'>, value: [{'type': 'text',' text': ' 4', 'index': 0}]
```

### Description

The `on_llm_new_token` callback method has a type hint of `token: str`, but when using Gemini 3 Flash (`gemini-3-flash-preview`), it receives a `list` of content blocks instead of a plain string.

This is a breaking change from Gemini 2.5 Flash behavior, where tokens were correctly passed as strings.

### Expected Behavior

The `langchain-google-genai` package should normalize token formats before calling `on_llm_new_token()`. Callbacks should receive plain strings regardless of which Gemini model version is used.

### Current Workaround

Handle multiple token formats in the callback:

```python
def on_llm_new_token(self, token: str, **kwargs) -> None:
    text = ""
    if isinstance(token, str):
        text = token
    elif isinstance(token, list):
        for block in token:
            if isinstance(block, dict) and "text" in block:
                text += block["text"]
    # Use text...
```

### Suggested Fix

In `langchain-google-genai`, extract text from content blocks before calling the callback:

```python
# Before calling on_llm_new_token:
if isinstance(token, list):
    token = "".join(block.get("text", "") for block in token if isinstance(block, dict))
```

### System Info

```
langchain-google-genai==2.1.4
langchain-core==0.3.62
google-generativeai==0.8.5
Python 3.13
```

Model: `gemini-3-flash-preview` (also affects `gemini-2.5-flash` in some configurations)

SanketJadhav7d3:
## Description

This PR fixes an inconsistency in streaming token handling for Gemini models in `langchain-google-genai`.

When using `gemini-3-flash-preview`, the `on_llm_new_token` callback receives a **list of content blocks** instead of a plain string, which breaks the expected callback contract (`token: str`). This differs from the behavior of `gemini-2.5-flash`, where tokens are emitted as strings.

This change normalizes streamed tokens by extracting and concatenating text from content blocks before invoking `on_llm_new_token`, ensuring consistent behavior across Gemini model versions.

## Relevant issues

Fixes #1460 

## Type

🐛 Bug Fix

## Changes(optional)

- Normalize streamed token payloads to plain strings before invoking `on_llm_new_token`
- Ensure consistent callback behavior across Gemini 2.5 Flash and Gemini 3 Flash Preview

## Example Usuage

Manually verified streaming behavior for both Gemini 2.5 Flash and Gemini 3 Flash Preview.

```python
from langchain_core.callbacks import BaseCallbackHandler

class StreamingCallback(BaseCallbackHandler):
    def on_llm_new_token(self, token: str, **kwargs) -> None:
        print(f"Token type: {type(token)}, value: {token}")

llm = ChatGoogleGenerativeAI(
    model="gemini-3-flash-preview",
    streaming=True,
    callbacks=[StreamingCallback()]
)

response = llm.invoke("Hi?")
```

### Output 

```bash
Token type: <class 'str'>, value: Hello! How can I help you today?
Token type: <class 'str'>, value:
Token type: <class 'str'>, value:
```

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -3138,23 +3138,19 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
                 gen = cast("ChatGenerationChunk", _chat_result.generations[0])
                 message = cast("AIMessageChunk", gen.message)
 
-                # normalize if instance of list
+                # Extract text for callback token parameter.
+                # We concatenate all text blocks but preserve the original gen/message
+                # so that non-text blocks (thinking, code execution, etc.) remain
+                # accessible via chunk.message.content_blocks in callbacks.
                 if isinstance(message.content, list):
-                    text_content = "".join(
-                        item.get("text", "")
-                        for item in message.content
-                        if isinstance(item, dict) and item.get("type") == "text"
+                    token_text = "".join(
+                        block.get("text", "")
+                        for block in message.content_blocks
+                        if isinstance(block, dict) and block.get("type") == "text"
                     )
-                    normalized_message = AIMessageChunk(
-                        content=text_content,
-                        additional_kwargs=message.additional_kwargs,
-                        response_metadata=message.response_metadata,
-                        tool_call_chunks=message.tool_call_chunks,
-                        id=message.id,
-                    )
-                    gen = ChatGenerationChunk(
-                        message=normalized_message,
-                        generation_info=gen.generation_info,
+                else:
+                    token_text = (
+                        message.content if isinstance(message.content, str) else ""
                     )
 
             # populate index if missing
@@ -3174,7 +3170,7 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
             )
 
             if run_manager:
-                run_manager.on_llm_new_token(gen.text, chunk=gen)
+                run_manager.on_llm_new_token(token_text, chunk=gen)
             yield gen
 
     async def _astream(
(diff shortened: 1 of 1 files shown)


mdrxy:
Thanks for the report and investigation. This issue is being addressed in `langchain-core` instead, which is the correct layer for this fix.

The root cause is in `BaseChatModel.stream()` in `langchain-core`, which passes `chunk.message.content` directly to `on_llm_new_token`. When content is a list of blocks (as with Gemini 3's thought signatures), this violates the `token: str` contract.

Fixing this in the integration layer is a workaround that:

1. Only benefits this one integration, not others that may encounter similar issues
2. Requires each integration to implement the same normalization logic
3. Doesn't address the actual bug in BaseChatModel
4. Would require breaking Gemini 3 `content` to become a `str`, losing required signatures

See https://github.com/langchain-ai/langchain/pull/34727

mdrxy:
Closing the loop here: this was fixed upstream in `langchain-core` by https://github.com/langchain-ai/langchain/pull/34739 and released in `langchain-core>=1.4.4`.

Rather than normalizing Gemini streamed content blocks to strings inside `langchain-google-genai`, the shared callback contract now supports structured content-block tokens directly. That avoids duplicating provider-specific normalization and preserves metadata such as Gemini thought signatures.

Thanks again for the investigation and proposed fix.

PR #1518, opened 2026-01-16
feat(genai): route URL inputs to file_data

junaidaslam2006:
Summary
Enable expanded inputs in GenAI by routing HTTP/HTTPS and GCS URIs directly to file_data instead of downloading content on the client side.

Motivation
Previously, URL-based inputs required downloading the content before being sent to GenAI, which increased latency and client-side overhead. This change allows supported URIs to be passed as file references, improving performance and simplifying input handling.

Changes

* Updated _convert_to_parts and ImageBytesLoader to pass HTTP/HTTPS and GCS URLs via file_data
* Added automatic MIME type inference for URL-based inputs
* Enabled URL support in standard content blocks (data, image_url) using file references
* Updated unit tests to validate correct file_data mapping for URL inputs

Testing

* Added and updated GenAI unit tests for URL and file-based parts
* Verified that URL inputs are correctly routed without client-side downloads

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -1073,15 +1073,16 @@ def _parse_response_candidate(
                 outcome = 1
             else:
                 outcome = 2
+            execution_result: dict[str, Any]
             if output_version == "v1":
-                execution_result: dict[str, Any] = {
+                execution_result = {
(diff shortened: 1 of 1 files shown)


junaidaslam2006:
@lkuligin I ran the tests locally and the issue is fixed; let me know if you need any further changes.

will be waiitng your review on all my 3 PRs

PR #1581, opened 2026-02-11
fix(genai): update dummy thought signature to bypass string (fixes #1570)

### LINKED ISSUE #1570: DUMMY_THOUGHT_SIGNATURE fallback rejected by Vertex AI — base64 encoding prevents bypass string from reaching the API
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
from google.genai._common import encode_unserializable_types
from google.genai.types import Part, FunctionCall
from langchain_google_genai.chat_models import DUMMY_THOUGHT_SIGNATURE
import json

# 1. Show what DUMMY_THOUGHT_SIGNATURE is
print(f"DUMMY type: {type(DUMMY_THOUGHT_SIGNATURE)}")
print(f"DUMMY length: {len(DUMMY_THOUGHT_SIGNATURE)} bytes")

# 2. Create a Part the way the library does it (function call + DUMMY signature)
part = Part(
    function_call=FunctionCall(name="search_companies", args={"name": "Acme"}),
    thought_signature=DUMMY_THOUGHT_SIGNATURE,
)

# 3. SDK serialization base64-encodes the bytes — API rejects this
serialized = part.model_dump(exclude_none=True)
encoded = encode_unserializable_types(serialized)
print(f"\nDUMMY after SDK serialization: {json.dumps(encoded['thought_signature'][:60])}...")
print("^ base64 blob — API rejects with 'Thought signature is not valid'")

# 4. The documented bypass string also gets corrupted
part2 = Part(
    function_call=FunctionCall(name="search_companies", args={"name": "Acme"}),
    thought_signature=b"skip_thought_signature_validator",
)
serialized2 = part2.model_dump(exclude_none=True)
encoded2 = encode_unserializable_types(serialized2)
print(f"\nBypass bytes after SDK serialization: {json.dumps(encoded2['thought_signature'])}")
print("^ base64-encoded — API does NOT recognize this as the bypass")

# 5. What the API actually needs in the JSON payload
print(f"\nWhat API expects: \"skip_thought_signature_validator\" (literal string)")
```

```shell
langchain_google_genai.chat_models.ChatGoogleGenerativeAIError: Error calling model 'gemini-3-pro-preview' (INVALID_ARGUMENT): 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'Unable to submit request because function call `hubspot_mcp_search_companies` in the 2. content block is missing a `thought_signature`. Learn more: https://docs.cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures', 'status': 'INVALID_ARGUMENT'}}
```

### Description

When using `ChatGoogleGenerativeAI` with Gemini 3+ models in a multi-turn tool-calling loop (e.g. `create_react_agent`), the library's `DUMMY_THOUGHT_SIGNATURE` fallback is **always rejected** by the API.

#### Root Cause

The `DUMMY_THOUGHT_SIGNATURE` constant ([chat_models.py:571-578](https://github.com/langchain-ai/langchain-google/blob/main/libs/genai/langchain_google_genai/chat_models.py#L571-L578)) is 311 bytes captured from a previous API response. It's injected into `Part.thought_signature` at [line 799](https://github.com/langchain-ai/langchain-google/blob/main/libs/genai/langchain_google_genai/chat_models.py#L799) when a function-call Part lacks a signature.

However, the Google GenAI SDK's `encode_unserializable_types()` in `google.genai._common` **base64-encodes all `bytes` values** before JSON serialization. The API receives a base64 blob instead of a valid thought signature:

```
Part(thought_signature=DUMMY_THOUGHT_SIGNATURE)   # 311 bytes
  → convert_to_dict()                             # bytes stay as bytes
  → encode_unserializable_types()                 # base64.urlsafe_b64encode(bytes) → string
  → json.dumps()                                  # "thoughtSignature": "<base64 blob>"
  → API rejects: "Thought signature is not valid"
```

The [documented bypass](https://cloud.google.com/vertex-ai/generative-ai/docs/thought-signatures) is the **literal string** `"skip_thought_signature_validator"` in the JSON `thoughtSignature` field. But since `Part.thought_signature` is typed as `Optional[bytes]`, there is no way to pass this string through the SDK's serialization pipeline without it being base64-encoded into something else.

#### When This Happens

In a ReAct agent loop, the first LLM call produces an AIMessage with tool calls and ephemeral thought signatures. On the second LLM call, this message is replayed in history. The original signatures are stale (tied to the first response context). The library's `_parse_chat_history` detects missing/stale signatures and injects `DUMMY_THOUGHT_SIGNATURE`, but the SDK corrupts it via base64 encoding before it reaches the API.

#### Observed Errors

Depending on what's sent in the `thoughtSignature` field, the API returns different errors:

- **Real (stale) signatures replayed** → `"Thought signature is not valid"`
- **`DUMMY_THOUGHT_SIGNATURE` bytes** (base64-encoded by SDK) → `"Thought signature is not valid"`
- **No signature at all** (`None`) → `"function call is missing a thought_signature"`
- **`b"skip_thought_signature_validator"`** in `Part.thought_signature` → `"Thought signature is not valid"` (SDK base64-encodes it to `"c2tpcF90aG91Z2h0X3NpZ25hdHVyZV92YWxpZGF0b3I="`)

**None of these work.** The only thing the API accepts is the literal string `"skip_thought_signature_validator"` in the JSON payload, which is impossible to achieve through the current `Part.thought_signature` → SDK serialization path.
on path.

### Example Code (Python)

```python
"""
Demonstrates that DUMMY_THOUGHT_SIGNATURE gets base64-encoded by the SDK,
making it impossible for the API to accept it.
"""

from google.genai._common import encode_unserializable_types
from google.genai.types import Part, FunctionCall
from langchain_google_genai.chat_models import DUMMY_THOUGHT_SIGNATURE
import json

# 1. Show what DUMMY_THOUGHT_SIGNATURE is
print(f"DUMMY type: {type(DUMMY_THOUGHT_SIGNATURE)}")
print(f"DUMMY length: {len(DUMMY_THOUGHT_SIGNATURE)} bytes")

# 2. Create a Part with a function call + DUMMY signature (what the library does)
part = Part(
    function_call=FunctionCall(name="search_companies", args={"name": "Acme"}),
    thought_signature=DUMMY_THOUGHT_SIGNATURE,
)

# 3. Show how the SDK serializes it — base64 encodes the bytes
serialized = part.model_dump(exclude_none=True)
encoded = encode_unserializable_types(serialized)
print(f"\nSerialized thoughtSignature: {json.dumps(encoded['thought_signature'][:60])}...")
print("^ This is a base64 blob — NOT a valid thought signature")

# 4. Show the bypass string also gets corrupted
part_bypass = Part(
    function_call=FunctionCall(name="search_companies", args={"name": "Acme"}),
    thought_signature=b"skip_thought_signature_validator",
)
serialized_bypass = part_bypass.model_dump(exclude_none=True)
encoded_bypass = encode_unserializable_types(serialized_bypass)
print(f"\nBypass as bytes → serialized: {json.dumps(encoded_bypass['thought_signature'])}")
print("^ SDK base64-encoded it — API will NOT recognize this as the bypass string")

# 5. What the API actually needs
print(f"\nWhat the API expects: \"skip_thought_signature_validator\"")
print("^ Literal string, not base64-encoded bytes")
```

**Output:**
```
DUMMY type: <class 'bytes'>
DUMMY length: 311 bytes

Serialized thoughtSignature: "ErQCCrECAdHtim8MtxgeMCRCiNiyoyImxtYAEDzz4NXOr_HSL3rA"...
^ This is a base64 blob — NOT a valid thought signature

Bypass as bytes → serialized: "c2tpcF90aG91Z2h0X3NpZ25hdHVyZV92YWxpZGF0b3I="
^ SDK base64-encoded it — API will NOT recognize this as the bypass string

What the API expects: "skip_thought_signature_validator"
^ Literal string, not base64-encoded bytes
```

### Suggested Fix

In `_parse_chat_history`, when injecting the fallback signature for missing thought signatures, the value needs to arrive as the **literal string** `"skip_thought_signature_validator"` in the JSON payload. Possible approaches:

1. **Coordinate with `google-genai` SDK** to special-case `thought_signature` in `encode_unserializable_types` (or accept `str` in the field).
2. **Post-process the request dict** after `encode_unserializable_types` runs but before `json.dumps`, replacing any base64-encoded `thought_signature` with the bypass string.
3. **Use a custom type** that survives serialization as a plain string.

### Workaround

We monkey-patch `google.genai._common.encode_unserializable_types` to intercept the `thought_signature` key and replace the bytes with the literal bypass string:

```python
import google.genai._common as _common_mod

_original_encode = _common_mod.encode_unserializable_types

def _patched_encode(data: dict[str, object]) -> dict[str, object]:
    result = _original_encode(data)
    if isinstance(result, dict) and "thought_signature" in result:
        result["thought_signature"] = "skip_thought_signature_validator"
    return result

_common_mod.encode_unserializable_types = _patched_encode
```

### Environment

- `langchain-google-genai`: 4.2.0
- `google-genai`: 1.60.0
- `langchain-core`: 1.2.7
- `langgraph`: 1.0.7
- Python: 3.13.0

### Related Issues

- #1389 — `"skip_thought_signature_validator"` flag not supported by VertexAI Client (closed — workaround only helps manually constructed messages, not library-injected DUMMY)
- #1364 — Function call missing thought_signature with Gemini 3 and create_react_agent (closed as resolved by DUMMY patching, but DUMMY itself is rejected)
- #1431 — Image part missing thought_signature (closed)

ADITYAKUMARRAI2007:
**Description**
This PR fixes Issue #1570 where the `DUMMY_THOUGHT_SIGNATURE` fallback was being rejected by the Vertex AI API with an `INVALID_ARGUMENT` error.

**The Fix**
Updated `DUMMY_THOUGHT_SIGNATURE` to use the documented bypass bytes `b"skip_thought_signature_validator"`. This ensures that when the SDK serializes the signature, the API recognizes the bypass intent instead of seeing an invalid base64 blob.

**Issue:** Fixes #1570


peterkarman1:
Hi, checking in on this PR. We reproduced the issue fixed here with `langchain-google-genai==4.2.2`, `gemini-3.1-flash-lite-preview`, and an active-loop tool-call history whose `AIMessage` did not originate from Gemini. The library injects `DUMMY_THOUGHT_SIGNATURE`, and the live Vertex call fails with `Thought signature is not valid.` Is this PR still the intended fix path for #1570, and is there anything needed to move it forward?

mdrxy:
Thanks for getting this started! We're superseding this with a stacked PR pair — the first adds a failing test case (#1971), and the second (#1972) addresses both the constant swap and the SDK serialization issue (which this PR didn't cover — `encode_unserializable_types` base64-encodes the bypass bytes before the request is sent, so the API still wouldn't see the literal string). Linking back here for context.

PR #1587, opened 2026-02-13
fix(genai): handle empty items dict in array schema conversion

### LINKED ISSUE #1768: Tools with List[Any] parameters fail with 400 errors
### Bug
In `ChatGoogleGenerativeAI` invocations, tools with `List[Any]` parameters fail at runtime with:

```
400 INVALID_ARGUMENT: GenerateContentRequest.tools[N].function_declarations[N].parameters.properties[field].items: missing field
```

`ChatVertexAI` works.

### Example
```
from typing import Any, List

@tool
def my_tool(data: List[Any]) -> str:
    """Tool with a List[Any] parameter."""
    return str(data)

# enter custom args for authentication
args = {}
for name, llm in [
    ("ChatGoogleGenerativeAI", ChatGoogleGenerativeAI(model="gemini-3.1-flash-lite-preview", **args),
    ("ChatVertexAI", ChatVertexAI(model="gemini-3.1-flash-lite-preview", **args),
]:
    try:
        llm.bind_tools([my_tool], tool_choice="my_tool").invoke("call my_tool")
        print(f"{name}: OK")
    except Exception as e:
        print(f"{name}: {e}")

#### Output:

ChatGoogleGenerativeAI: Error calling model 'gemini-3.1-flash-lite-preview' (INVALID_ARGUMENT): 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': '*GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[data].items: missing field.\n', 'status': 'INVALID_ARGUMENT'}}
ChatVertexAI: OK
```

mdrxy:
Fixes #1768

---

Fix silent data loss when converting tool schemas with empty `items` dicts (`"items": {}`) to Gemini API format. 

When a tool parameter is typed as `list[Any]` or `list[list[Any]]`, Pydantic emits `{"type": "array", "items": {}}`. Because `bool({})` is `False` in Python, two separate truthiness checks caused the `items` field to be silently dropped. The Gemini API then rejects these schemas with `INVALID_ARGUMENT` ("missing field").

---

> ["In JSON Schema, an empty object is a completely valid schema that will accept any valid JSON."](https://json-schema.org/understanding-json-schema/basics)


lkuligin (dismissed):

libs/genai/langchain_google_genai/_function_utils.py
```diff
@@ -130,7 +130,7 @@ def _dict_to_genai_schema(
     is_property: bool = False,
     is_any_of_item: bool = False,
 ) -> types.Schema | None:
```
lkuligin, line 132: ideally, we should use a dict here and pass it as `parameters_json_schema`

PR #1588, opened 2026-02-14
fix(genai): propagate model-level stop sequences to generation config

### LINKED ISSUE #1586: [ChatGoogleGenerativeAI] Model attribute `stop` is not propagated to the model
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
from langchain_google_genai.chat_models import ChatGoogleGenerativeAI

llm = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash-lite",
    stop=["model"],
)

# This gives the wrong output
llm.invoke("Repeat after me: 'I am a model from Google.'")

# This works correctly
llm.invoke("Repeat after me: 'I am a model from Google.'", stop=["model"])
```

```shell
No error, but output is not as expected.

AIMessage(content='I am a model from Google.', ...)
```

### Description

1. Create a model using stop sequences. (`.stop` is a [documented model parameter](https://reference.langchain.com/python/integrations/langchain_google_genai/ChatGoogleGenerativeAI/#langchain_google_genai.ChatGoogleGenerativeAI.stop))
2. Expected the output to stop short of model: "I am a".
3. Got "I am a model from Google."

It works correctly when we specify the sequence in the `invoke()` method, but not when we do it on the model level.

The described expected behaviour would be consistent with the behaviour of the Google Genai SDK's Chat:

```python
from google import genai
client = genai.Client()
chat = client.chats.create(
    model="gemini-2.5-flash-lite",
    config=genai.types.GenerateContentConfig(
        stop_sequences=["model"],
    ),
)
response = chat.send_message("Are you an LLM?")
response.text

>>> 'I am a large language '
```

And with some random other model through LangChain:

```python
from langchain_openai import ChatOpenAI
llm = ChatOpenAI(
    model="gpt-4.1-nano",
    stop_sequences=["model"],
)
llm.invoke("Are you an LLM?")

>>> AIMessage(content='Yes, I am a large language ', ...)
```

Similar to #1454 but not the exact same cause.

I think I found the root cause and a fix, but it needs some verification and testing. Then I'll contribute.

junaidaslam2006:
## Description
Fixes #1586 — Model attribute `stop` is not propagated to the generation config.

When `stop` is set as a model attribute on `ChatGoogleGenerativeAI`, it was silently ignored unless also provided at call time via `invoke(stop=...)`. This is because `_build_base_generation_config` used the `stop` parameter directly without falling back to `self.stop`.

## Fix
Changed `"stop_sequences": stop` to `"stop_sequences": stop if stop is not None else self.stop`, consistent with how other parameters (temperature, max_output_tokens, etc.) are already handled.

## Tests
Added `test_model_level_stop_propagated` unit test verifying that model-level stop sequences appear in the generation config when no override is passed at call time.


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -2584,7 +2582,7 @@ def _build_base_generation_config(
         config: dict[str, Any] = {
             "candidate_count": self.n,
             "temperature": kwargs.get("temperature", self.temperature),
-            "stop_sequences": stop,
+            "stop_sequences": stop if stop is not None else self.stop,
```
julesvanrie, line 2585: LGTM: This is how I solved it too.

libs/genai/tests/integration_tests/test_chat_models.py
```diff
@@ -2518,8 +2518,14 @@ def test_context_caching(backend_config: dict) -> None:
...
-    assert isinstance(response.content, str)
-    assert "747" in response.content
+    content = response.content
+    if isinstance(content, list):
+        text = " ".join(
+            b["text"] for b in content if isinstance(b, dict) and b.get("text")
+        )
+    else:
+        text = content
+    assert "747" in text
```
julesvanrie, line 2528: These changes (as well as the next) are unrelated to the issue, but to pass the integration tests, right? I saw you added the same in #1585.

julesvanrie (commented):
After submitting the issue I started working on a solution, but then I saw your PR. So I dropped my work because I was essentially writing the same solution.

libs/genai/tests/integration_tests/test_chat_models.py
junaidaslam2006: they're required to make CI pass so I included them here to keep the PR mergeable and also in #1585

lkuligin (approved):

PR #1590, opened 2026-02-16
fix(vertexai): use model-aware `max_output_tokens` defaults for `ChatAnthropicVertex`

### LINKED ISSUE #1584: ChatAnthropicVertex hardcodes max_output_tokens=1024, should use model-aware defaults like ChatAnthropic
### Package

### Checked other resources

### Example Code (Python)

```python
from langchain_anthropic import ChatAnthropic

# ChatAnthropicVertex defaults to 1024
vertex_model = ChatAnthropicVertex(model_name="claude-sonnet-4-5", project="my-project", location="us-east5")
print(vertex_model.max_output_tokens)  # 1024

# ChatAnthropic uses model-aware defaults
direct_model = ChatAnthropic(model="claude-sonnet-4-5")
print(direct_model.max_tokens)  # 64000
```

No error — outputs are silently truncated to 1024 tokens. When using `with_structured_output()`, the JSON response is cut off mid-generation and the output parser receives incomplete data, resulting in empty or malformed results.

### Description

`ChatAnthropicVertex` defines `max_output_tokens` with a hardcoded default of 1024 in `model_garden.py`:

```python
max_output_tokens: int = Field(default=1024, alias="max_tokens")
```

Meanwhile, `ChatAnthropic` uses a model-aware validator that looks up per-model defaults from `_MODEL_DEFAULT_MAX_OUTPUT_TOKENS` — for example, `claude-sonnet-4-5` defaults to 64000 and `claude-opus-4-5` defaults to 32000.

This discrepancy means that switching from `ChatAnthropic` to `ChatAnthropicVertex` silently truncates outputs to 1024 tokens unless the caller explicitly sets `max_output_tokens`. This is especially problematic with structured output / tool use, where the model's JSON response gets cut off mid-generation.

The fix would be to apply the same model-aware default logic in `ChatAnthropicVertex` that `ChatAnthropic` already uses — looking up the model name against known max output token limits and falling back to a sensible default (e.g. 4096).

Deeven-Seru:
## Summary

ChatAnthropicVertex previously hardcoded `max_output_tokens` to 1024, which silently truncated outputs for modern Claude models that support much higher limits. This was especially problematic with `with_structured_output()` where the JSON response would be cut off mid-generation, producing empty or malformed results.

## Root Cause

The field definition used a static default:
```python
max_output_tokens: int = Field(default=1024, alias="max_tokens")
```

Meanwhile, `ChatAnthropic` from `langchain-anthropic` uses model-aware defaults (e.g. `claude-sonnet-4-5` -> 64000, `claude-opus-4-0` -> 32000). Users switching from `ChatAnthropic` to `ChatAnthropicVertex` would silently get truncated outputs.

## Changes

- Added a lookup table of known Anthropic model max output token limits (`_ANTHROPIC_MODEL_MAX_OUTPUT_TOKENS`)
- Added a `model_validator(mode="before")` that sets model-aware defaults when `max_output_tokens` is not explicitly provided by the user
- Falls back to 4096 for unknown models (consistent with `langchain-anthropic`)
- Explicit user-provided values are never overridden

## Tests

Added 4 new test functions covering:
- Parametrized model-aware defaults for 9 different models (claude-sonnet-4-5, claude-opus-4-0, claude-opus-4-6, claude-3-5-sonnet, claude-3-7-sonnet, etc.)
- Unknown model fallback to 4096
- Explicit `max_output_tokens` override preserved
- Explicit `max_tokens` alias override preserved

All existing tests continue to pass.

Fixes #1584


libs/vertexai/langchain_google_vertexai/model_garden.py
```diff
@@ -154,10 +154,65 @@ async def _agenerate(
...
+        model_name, _ANTHROPIC_FALLBACK_MAX_OUTPUT_TOKENS
+    )
+
+
 class ChatAnthropicVertex(_VertexAICommon, BaseChatModel):
     async_client: Any = Field(default=None, exclude=True)
 
-    max_output_tokens: int = Field(default=1024, alias="max_tokens")
+    max_output_tokens: int = Field(
+        default=_ANTHROPIC_FALLBACK_MAX_OUTPUT_TOKENS, alias="max_tokens"
```
lkuligin: this is backwards incompatible change, why would we need it? and it might inflate the costs and change the model's behaviour.
if we feel like users necessarily need it, let's expose `_get_anthropic_max_output_tokens` and users can use it if they wish, but I'd be against changing the defaults.

lkuligin (changes requested):

libs/vertexai/langchain_google_vertexai/model_garden.py
mdrxy: Thanks for the review. Pushing back on this:

**The 1024 default is a bug, not an interface contract.** It silently truncates on every modern Claude model — `with_structured_output()` is unusable beyond a few hundred tokens because partial JSON doesn't parse (#1584). Explicit `max_output_tokens=N` / `max_tokens=N` is preserved untouched, so callers with intentional caps are unaffected.

**Cost is not a real concern here.** `max_tokens` is a cap, not a target — billing is per generated token. Raising the cap only changes cost when the model would have been cut off at 1024, in which case the prior response was broken. Users wanting tight cost caps should set `max_tokens` explicitly; relying on a magic 1024 default for cost control isn't reliable.

**Parity with `langchain-anthropic` is the bar.** `ChatAnthropic` already defaults to model-profile values (4096 fallback, per-model overrides). Switching between `ChatAnthropic` and `ChatAnthropicVertex` should not silently degrade output limits.

The latest revision pulls profile data through the same `langchain-profiles` pipeline `langchain-anthropic` uses (`libs/vertexai/langchain_google_vertexai/data/anthropic/_profiles.py`, regenerated from models.dev). Default stays at 4096; per-model profile only overrides via the validator at construction. Same mechanism, same data source — pivoting to opt-in would just preserve the bug.

PR #1593, opened 2026-02-17
feat(genai): `tools` on `with_structured_output`

mdrxy:
Add a first-class `tools` parameter to `ChatGoogleGenerativeAI.with_structured_output` so users can combine structured output with Google Search (or other grounding tools) in a single call. Previously this required a manual `.bind()` workaround with three separate kwargs — now it's a one-liner on the `json_schema` path.

This will require a docs update.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -3268,6 +3268,10 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
         | None = "json_schema",
         *,
         include_raw: bool = False,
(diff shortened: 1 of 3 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -3268,6 +3268,10 @@ def with_structured_output(
         | None = "json_schema",
         *,
         include_raw: bool = False,
+        tools: Sequence[
+            dict[str, Any] | type | Callable[..., Any] | BaseTool | GoogleTool
```
ccurme, line 3272: This is typed as accepting BaseTool, but we don't appear to convert to provider format.

PR #1602, opened 2026-02-21
fix(community): resolve async bigquery test leak by using generator fixtures

saravana-tk:
## Description
While running the unit tests for `langchain-google-community`, I noticed that `test_bigquery_callbacks.py` was throwing several `Task was destroyed but it is pending!` errors from `asyncio` at the end of the test suite. 

The issue was caused by the `handler` and `sync_handler` fixtures using `return` instead of `yield`. Because `_ensure_started()` spins up a background batch-writing task, the tests were finishing and closing the event loop before the background tasks were cleanly shut down.

## Relevant issues
N/A

## Type
🐛 Bug Fix
✅ Test

## Changes
* Converted `handler` to an `AsyncGenerator` and `sync_handler` to a `Generator`.
* Replaced `return handler` with `yield handler` in the fixtures.
* Added `await handler.shutdown()` and `handler.shutdown()` respectively to the teardown phase to cleanly close the background workers and prevent the memory leak. 
* Updated typing imports to include `AsyncGenerator`.

## Testing
Ran `make test` locally in the `libs/community` directory. The test suite passes cleanly, and the dangling `asyncio` background task errors are completely resolved.

## Note
Because the fixtures now properly wait for the background tasks to shut down before yielding control back to Pytest, the test suite is slightly slower, but entirely safe from background memory leaks.


saravana-tk:
Looks like the llm-integration-tests job failed on Google Cloud Build, but all the Python unit tests for make test passed successfully. Since my changes were isolated to the tests/unit_tests/ directory, I suspect this might be a flaky integration test or a timeout. Let me know if you need me to do anything on my end

saravana-tk:
"cc: @lkuligin - Tagging you as I saw you actively reviewing the Google GenAI packages. The branch is synced, CI is fully green, and the asyncio memory leaks in the BigQuery callbacks are resolved. This PR is ready for a quick review whenever you have time."

Issue #1608, opened 2026-02-24
Bug: `with_structured_output()` returns empty nested lists on Vertex AI due to responseJsonSchema


phamgialinhlx:
### Package (Required)

- [x] langchain-google-genai
- [ ] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
"""Fuller reproduction: with_structured_output() + nested schema on Vertex AI.

Uses the richer schema and larger dataset from the bug report to trigger
the empty-array behavior on Vertex AI.
"""

from pydantic import BaseModel, Field
from langchain_google_genai import ChatGoogleGenerativeAI
from langchain_core.messages import HumanMessage, SystemMessage


class Product(BaseModel):
    """A single product entry in a store inventory."""
    name: str = Field(..., description="Product name")
    price: str = Field("", description="Price as displayed (e.g. '$4.99')")
    quantity: str = Field("", description="Stock quantity or amount")
    unit: str = Field("", description="Unit of measure (e.g. kg, each, pack)")
    category: str = Field("", description="Product category or aisle")
    notes: str = Field("", description="Additional info (e.g. 'on sale', 'organic')")


class StoreInventory(BaseModel):
    """Store inventory listing with nested product items."""
    store_name: str = Field("", description="Name of the store")
    store_location: str = Field("", description="Store address or location")
    items: list[Product] = Field(default_factory=list, description="List of products in inventory")
    notes: str = Field("", description="Additional notes about the inventory")


INVENTORY_TEXT = """\
Extract ALL products from this store inventory.

GREENFIELD GROCERY — MAIN STREET BRANCH
Weekly Inventory Report

PRODUCE
Organic Bananas       $1.29/lb      150 lbs    [sale]
Red Apples            $2.49/lb      200 lbs
Baby Spinach          $3.99/bag     85 bags    [organic]
Avocados              $1.50/each    120 each
Cherry Tomatoes       $4.29/pint    60 pints
Russet Potatoes       $0.99/lb      300 lbs
Yellow Onions         $1.19/lb      175 lbs
Fresh Basil           $2.99/bunch   40 bunches
Carrots               $1.79/lb      130 lbs
[...]

PR #1611, opened 2026-02-24
fix(genai): route response_schema directly to GenerateContentConfig

phamgialinhlx:
## Description

When using `with_structured_output` with Pydantic models that produce `$defs`/`$ref` in their JSON schema (e.g., models with `Optional` fields, `Union` types, or recursive structures), Vertex AI silently
returns empty arrays instead of valid structured output.

The root cause is that `response_schema` was being passed through `GenerationConfig`, whose strict `Schema` type rejects `$defs`/`$ref` fields. This fix routes `response_schema` directly to
`GenerateContentConfig`, bypassing `GenerationConfig`, so the SDK's `process_schema()` pipeline can inline `$defs`/`$ref` before sending the request to Vertex AI.

AI agents were used to assist with this contribution.

## Relevant issues
- https://github.com/langchain-ai/langchain-google/issues/1608

## Type

🐛 Bug Fix

## Changes(optional)

- `with_structured_output` now binds via `response_schema` instead of `response_json_schema`
- `_prepare_request` extracts `response_schema` from kwargs before `_prepare_params` and passes it directly to `_build_request_config`
- `_add_response_params` no longer handles `response_schema` (only `response_json_schema`)
- `_build_request_config` accepts an optional `response_schema` and forwards it to `GenerateContentConfig`

## Testing(optional)

Updated unit tests in `test_chat_models.py`:
- `test_response_json_schema_param_mapping`: verifies `response_schema` bypasses `GenerationConfig` and appears on `GenerateContentConfig`
- `test_ref_preservation`: updated to check `response_schema` key
- `test_union_schema_support`: updated to check `response_schema` key
- `test_response_schema_mime_type_validation`: validation now tested via `_prepare_request`

## Note(optional)

This is backwards compatible. The `response_json_schema` kwarg continues to work as before through `GenerationConfig`. Only `response_schema` and the `with_structured_output` code path are affected.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -2824,27 +2824,22 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
     ) -> dict[str, Any]:
         """Add response-specific parameters to generation config.
 
-        Includes `response_mime_type`, `response_schema`, and `response_json_schema`.
+        Handles `response_mime_type` and `response_json_schema`.
+
+        Note: `response_schema` is handled separately in `_prepare_request`
+        because it must bypass GenerationConfig (whose strict Schema type
+        rejects $defs/$ref) and go directly to GenerateContentConfig.
         """
         # Handle response mime type
         response_mime_type = kwargs.get("response_mime_type", self.response_mime_type)
         if response_mime_type is not None:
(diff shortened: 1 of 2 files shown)


octaviogl:
When using VertexAI endpoint through "structured output" (`vertexai:true`), _response_json_schema_ parameter makes the model to build a wrong response, and even to have a poor reasoning. Allowing _response_schema_ to be used is a **must**.
Thank you!

PR #1530, opened 2026-01-21
fix(genai): guard cached_content with tools

junaidaslam2006:
## Summary
- **Preserve list-based content:** Updated for Gemini 3 and output version `v1` to prevent content from collapsing into strings.
- **Propagate `output_version`:** Handled via response parsing/metadata without sending it to the request config.
- **API Constraint Guard:** Prevent `cached_content` from being combined with `tools`, `tool_config`, or `system_instruction` to comply with Gemini API requirements.

Fixes #1528

## Testing
I have verified the fix using the following integration tests:

```bash
pytest libs/genai/tests/integration_tests/test_chat_models.py::test_search_builtin
pytest libs/genai/tests/integration_tests/test_chat_models.py::test_structured_output_with_google_search
pytest libs/genai/tests/integration_tests/test_chat_models.py::test_gemini_3_pro_agent_loop_streaming
pytest libs/genai/tests/integration_tests/test_standard.py::TestGeminiProStandardGoogleAI::test_structured_output_optional_param

PR #1619, opened 2026-03-01
fix(genai): strip tools/tool_config/system_instruction from request when cached_content is set

### LINKED ISSUE #1528: Gemini Context Caching Incompatibility with LangChain Tools/Agents
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
import os
import time
from google import genai
from google.genai import types


load_dotenv()


# 1. Setup Client & File
with open("temp_cache_doc.txt", "w") as f:
    f.write("This is a test document. " * 500) # Ensure ample tokens

client = genai.Client(api_key=os.environ["GOOGLE_API_KEY"])

uploaded_file = client.files.upload(file="temp_cache_doc.txt")

# 2. Create Cache (Content Only)
# We might hope to bind tools later via LangChain...
cache = client.caches.create(
    model="gemini-2.5-flash",
    config=types.CreateCachedContentConfig(
        display_name="Test Cache",
        contents=[
            types.Content(
                role="user",
                parts=[types.Part.from_uri(file_uri=uploaded_file.uri, mime_type=uploaded_file.mime_type)]
            )
        ],
        ttl="300s",
    ),
)

# 3. Initialize LangChain Model with Cache
llm = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash",
    cached_content=cache.name,
)

# 4. Define & Bind Tool
@tool
def get_weather(city: str):
    """Get weather for a city."""
    return "Sunny"

# ERROR HAPPENS HERE ON INVOKE
# bind_tools attaches the tool definition to the GenerateContent request
llm_with_tools = llm.bind_tools([get_weather])

try:
    response = llm_with_tools.invoke("What is the weather in Paris?")
    print(response)
except Exception as e:
    print("\nCAUGHT EXPECTED ERROR:")
    print(e)

# Cleanup
client.caches.delete(name=cache.name)
client.files.delete(name=uploaded_file.name)
```

```shell
Error calling model 'gemini-2.5-flash' (INVALID_ARGUMENT): 400 INVALID_ARGUMENT. {'error': {'code': 400, 'message': 'CachedContent can not be used with GenerateContent request setting system_instruction, tools or tool_config.\n\nProposed fix: move those values to CachedContent from GenerateContent request.', 'status': 'INVALID_ARGUMENT'}}
```

### Description

There is a fundamental incompatibility between **Google Gemini's Context Caching** feature and **LangChain's standard Agent/Tool binding mechanism**.

When using `cached_content` with `ChatGoogleGenerativeAI`, the Gemini API strictly forbids passing `tools` or `system_instruction` in the `GenerateContent` request. However, standard LangChain patterns (like `bind_tools` or `create_tool_calling_agent`) inherently inject these parameters into the generation request, causing an `INVALID_ARGUMENT` (400) error.

The Gemini API Error states:
> *CachedContent can not be used with GenerateContent request setting system_instruction, tools or tool_config.*

If you want to use tools or system instructions with a cache, they **must be defined within the cache configuration itself** at creation time.

However, `ChatGoogleGenerativeAI` and related agents do not currently support this workflow gracefully. When a user binds tools to the model (e.g., via `llm.bind_tools([...])`), LangChain sends them in the generation request standard payload. If `cached_content` is also provided in the model parameters, the API rejects the request.

### LINKED ISSUE #1618: fix(genai): strip tools/tool_config/system_instruction from request when cached_content is set — production-tested solution
## Related issues & PRs

- Closes #1528 — *CachedContent can not be used with GenerateContent request setting system_instruction, tools or tool_config*
- Supersedes / complements #1530 — *fix(genai): guard cached_content with tools* (see reasoning below)

---

## Why PR #1530's approach is incomplete

PR #1530 solves the error by raising a `ValueError` when `cached_content` is combined with `tools`, `tool_config`, or `system_instruction`. That protects users from accidentally sending an invalid request — but it **blocks the primary valid use case**: an agent that pre-bakes its tools and system instruction into the `CachedContent` at creation time and then reuses that cache across requests.

When tools are already encoded in the cache, they must **not** appear in the per-request payload — stripping them is correct behavior, not an error condition. Raising a `ValueError` here means nobody can use Gemini context caching with tool-calling agents at all, which defeats the purpose of the feature.

---

## The correct fix

When `cached_content` is set on a request, silently omit `tools`, `tool_config`, and `system_instruction` from `GenerateContentConfig` — because those values are already encoded in the cache:

```python
# In _build_request_config — after all formatting is done, before building GenerateContentConfig
if cached_content:
    formatted_tools = None
    formatted_tool_config = None
    system_instruction = None
```

This matches the Gemini API contract exactly: a request carrying `cached_content` must not repeat fields that are already stored in the cache.

---

## The full pattern we are offering

We've been running this in production across a multi-tenant LangGraph deployment using `ChatGoogleGenerativeAI` with per-role, per-scenario `CachedContent`. The pattern:

### 1 — Cache creation (once, TTL-controlled)

```python
from google.genai import types
from langchain_google_genai._function_utils import convert_to_genai_function_declarations

gemini_tools = convert_to_genai_function_declarations(langchain_tools)

cache = client.caches.create(
    model=model,
    config=types.CreateCachedContentConfig(
        display_name=f"agent-{role}-{scenario}",
        system_instruction=system_instruction,   # character prompt + scenario
        tools=gemini_tools,                      # all role tools
        ttl=f"{ttl_seconds}s",
    ),
)
```

### 2 — Per-request invocation

```python
llm = ChatGoogleGenerativeAI(model=model, cached_content=cache.name)
# tools are bound via .bind_tools() as usual for LangGraph agent nodes —
# _build_request_config must strip them before sending because they're in the cache
response = await llm.ainvoke(messages)
```

### 3 — The patch (what we monkey-patch today, should be native)

```python
# In ChatGoogleGenerativeAI._build_request_config:
if cached_content:
    formatted_tools = None
    formatted_tool_config = None
    system_instruction = None

return GenerateContentConfig(
    tools=list(formatted_tools) if formatted_tools else None,
    tool_config=formatted_tool_config,
    system_instruction=system_instruction,
    cached_content=cached_content,
    ...
)
```

---

## Production context

We've been running this pattern in production for several weeks across multiple tenants, multiple roles (Founder / Manager / Specialist), and multiple scenarios per role. Each scenario cache holds the full character prompt + scenario description + all role tools (~15–30 tools per role). The 75% token discount on cached tokens makes the pattern economically significant at scale.

The monkey-patch has been stable, but it's fragile — tied to a private method signature that could change across releases. We'd like to contribute this as a first-class fix so other teams don't have to patch their way to production.

---

## Proposed PR scope

1. **`_build_request_config` in `chat_models.py`**: strip `tools`, `tool_config`, `system_instruction` when `cached_content` is set
2. **Tests**: cover the case where `cached_content` is set and tools are bound — verify the request config omits the conflicting fields
3. **Docs/example** (optional): a minimal example showing the cache-creation + invocation pattern

Happy to open the PR directly. Flagging here first to align on approach before writing code.

Mukhsin0508:
## Description

When `cached_content` is set on a `ChatGoogleGenerativeAI` request, the Gemini API rejects any request that also includes `tools`, `tool_config`, or `system_instruction` — because those fields are already encoded inside the cache. This PR adds a guard in `_build_request_config` that strips those three fields from the per-request payload whenever `cached_content` is present.

This is the correct fix for the pattern where tools and the system instruction are **pre-baked into the `CachedContent` at creation time** and reused across requests. In that pattern, stripping them is correct behavior — not an error condition.

## Relevant issues

Fixes #1618
Closes #1528
See also #1530 (alternative approach that raises `ValueError` instead of stripping — blocks the valid agentic cache use case)

## Type

🐛 Bug Fix

## Changes

**`libs/genai/langchain_google_genai/chat_models.py`**

Added a guard at the top of `_build_request_config`:

```python
# When cached_content is set, tools/tool_config/system_instruction must
# not be included in the request — they are already encoded in the cache.
# https://github.com/langchain-ai/langchain-google/issues/1618
if cached_content:
    formatted_tools = None
    formatted_tool_config = None
    system_instruction = None
```

**`libs/genai/tests/unit_tests/test_chat_models.py`**

Two new unit tests (no network access):

- `test_cached_content_strips_tools_tool_config_system_instruction` — verifies that when `cached_content` is set, the resulting `GenerateContentConfig` has `tools=None`, `tool_config=None`, `system_instruction=None`
- `test_cached_content_not_set_preserves_tools_and_system_instruction` — verifies that when `cached_content` is not set, tools and system instruction remain in the config as expected

## Testing

```bash
cd libs/genai
make test
# or specifically:
pytest tests/unit_tests/test_chat_models.py \
  -k "test_cached_content"
```

Both tests pass. No network access required.

## Note

We've been running this exact fix as a monkey-patch in production for several weeks across a multi-tenant LangGraph deployment — multiple roles, multiple per-scenario caches, 15–30 tools per cache. Each cache holds the full system instruction + all role tools. The fix is stable and the 75% token discount on cached tokens is economically significant at scale.

The monkey-patch works but is fragile (tied to a private method signature). This PR makes it a first-class, tested behaviour so other teams building agentic systems on Gemini context caching don't have to patch their way to production.

### DIFF
--- a/libs/community/langchain_google_community/drive.py
+++ b/libs/community/langchain_google_community/drive.py
@@ -534,10 +534,39 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
 
     def _load_documents_from_ids(self) -> List[Document]:
         """Load documents from a list of IDs."""
+        from googleapiclient.discovery import build  # type: ignore[import]
+
         if not self.document_ids:
             raise ValueError("document_ids must be set")
 
-        return [self._load_document_from_id(doc_id) for doc_id in self.document_ids]
+        creds = self._load_credentials()
+        service = build("drive", "v3", credentials=creds)
+
+        q = " or ".join(f"id = '{doc_id}'" for doc_id in self.document_ids)
+        results = (
+            service.files()
+            .list(
+                q=q,
+                fields="files(id, mimeType)",
+                includeItemsFromAllDrives=True,
+                supportsAllDrives=True,
+            )
+            .execute()
+        )
+        mime_types: Dict[str, str] = {
+            f["id"]: f["mimeType"] for f in results.get("files", [])
+        }
+
+        documents: List[Document] = []
+        for doc_id in self.document_ids:
+            mime_type = mime_types.get(doc_id)
+            if mime_type == "application/vnd.google-apps.spreadsheet":
+                documents.extend(self._load_sheet_from_id(doc_id))
+            elif mime_type == "application/pdf" or self.file_loader_cls is not None:
+                documents.extend(self._load_file_from_id(doc_id))
+            else:
+                documents.append(self._load_document_from_id(doc_id))
+        return documents
 
     def _load_file_from_id(self, id: str) -> List[Document]:
         """Load a file from an ID."""
--- a/libs/community/tests/unit_tests/test_drive.py
+++ b/libs/community/tests/unit_tests/test_drive.py
@@ -1,4 +1,7 @@
+from unittest.mock import MagicMock, patch
+
 import pytest
+from langchain_core.documents import Document
 
 from langchain_google_community.drive import GoogleDriveLoader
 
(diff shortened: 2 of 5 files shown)


Mukhsin0508:
The CI failure (`test_context_caching[google_ai]`) is pre-existing and unrelated to this PR.

The build log confirms it runs against the upstream HEAD (`9226a247` — *"refactor: update type casting for tools in request preparation"*), not this PR's commit. The test fails because it asserts `isinstance(response.content, str)` but the model (a reasoning model, given `output_token_details: {reasoning: 200}`) now returns a list of structured content blocks with `extras.signature` fields — a model behaviour change on their end, not a regression introduced here.

Our unit tests (`test_cached_content_strips_tools_tool_config_system_instruction` and `test_cached_content_not_set_preserves_tools_and_system_instruction`) pass cleanly.

Mukhsin0508:
@mdrxy Hey!

Can you guys take a look, we want to clean our codebase from patches, this is holding us back.

Issue #1622, opened 2026-03-03
Nest schema field convention bug


jectpro7:
### Package (Required)

- [x] langchain-google-genai
- [ ] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
from pydantic import BaseModel, Field
from langchain_core.tools import tool
from langchain_google_genai._function_utils import (
    _format_base_tool_to_function_declaration,
)


class Step(BaseModel):
    name: str = Field(description="Step name")
    config: dict = Field(default={}, description="Arbitrary config dict")


class Plan(BaseModel):
    steps: list[Step] = Field(description="List of steps")


@tool
def create_plan(plan: Plan) -> str:
    """Create a plan."""
    return "ok"


fd = _format_base_tool_to_function_declaration(create_plan)
config_schema = (
    fd.parameters.properties["plan"]
    .properties["steps"]
    .items.properties["config"]
)
print(f"config field type: {config_schema.type}")
# Output: config field type: Type.STRING
# Expected: Type.OBJECT
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

Nested dict fields in tool schemas are silently converted to STRING type, causing Gemini to send JSON strings that fail Pydantic validation.

The LLM sees the field as a string, sends "{\"key\": \"value\"}", but the Python code expects a dict, resulting in a ValidationError: Input should be a valid dictionary.

PR #1625, opened 2026-03-05
fix(genai): preserve additionalProperties for nested freeform dict fi…

jectpro7:
## Description                                                                                          
              
  `_get_properties_from_schema` converts `OBJECT→STRING` when an object has no `properties` and no        
  `additionalProperties`. The output didn't preserve `additionalProperties`, so recursive re-processing
  during nested schema conversion would lose it and incorrectly fall back to STRING, causing Gemini to    
  send JSON strings for nested dict fields that fail Pydantic validation.                                 

  The fix preserves `additionalProperties` in intermediate output so it survives re-processing. This is
  safe because `_format_json_schema_to_gapic` strips it from the final schema.

  ## Type

  🐛 Bug Fix

  ## Changes

  - Preserve `additionalProperties` in `_get_properties_from_schema` output for OBJECT types
  - Add 9 tests covering low-level, mid-level, and end-to-end scenarios

  ## Testing

  ```bash
  cd libs/genai
  uv run --group test pytest tests/unit_tests/test_function_utils.py -x -v
  ```

### DIFF
--- a/libs/genai/tests/integration_tests/test_chat_models.py
+++ b/libs/genai/tests/integration_tests/test_chat_models.py
@@ -2518,8 +2518,15 @@ def test_context_caching(backend_config: dict) -> None:
     response = chat.invoke("What is the secret number?")
 
     assert isinstance(response, AIMessage)
-    assert isinstance(response.content, str)
-    assert "747" in response.content
+    assert isinstance(response.content, (str, list))
+    response_text = (
+        response.content
+        if isinstance(response.content, str)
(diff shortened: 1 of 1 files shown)


jectpro7:
fixed https://github.com/langchain-ai/langchain-google/issues/1622

jectpro7:
Hi @jkennedyvz and @lkuligin, could you please help review this PR when you get a chance? Thanks!

PR #1630, opened 2026-03-07
fix(vertexai): allow thought_signature key in _collapse_text_content

### LINKED ISSUE #1621: [VertexAI] response.content returns list instead of str when `thinking_budget` is set and `thought_signature` is present in text blocks
### Package (Required)

- [ ] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
import asyncio
from langchain_core.messages import HumanMessage, SystemMessage

@tool
def get_weather(city: str) -> str:
    """Get weather for a city."""
    return f"Weather in {city}: sunny"

async def main():
    model = ChatVertexAI(
        project="your-project-id",
        model="gemini-2.5-flash",
        thinking_budget=-1,  # enables dynamic thinking
    )

    response = await model.bind_tools([get_weather]).ainvoke([
        SystemMessage(content="You are helpful."),
        HumanMessage(content="Hello, how are you?")
    ])

    print(type(response.content))  # <class 'list'>, expected <class 'str'>
    print(response.content)
    # [{'type': 'text', 'text': '...', 'thought_signature': '...'}]

asyncio.run(main())
```

```shell
No exception is raised. The bug is a silent type mismatch:
response.content is <class 'list'> instead of <class 'str'>.

Content value:
[{'type': 'text', 'text': "Hello! I'm doing well, thank you for asking. How can I assist you today?", 'thought_signature': 'CqMBAY89a1+8HH8l+Dh/CYHi1cES97yA0qT96XbFIRRAuXNlSD3qM2WJpSe5sET+vlMeVtM14pObFxjoqWtriVOUdU0g8vHHI5u70ZkOAJtoBldIWYpAhrWJzyYNaTgyTrCHUMqp4kqvIBpdC2lN8mgegAoev29Kl0tTBKVruPPJEtgj8LeABo3GFt9kkojbeOEK+xSmBd90gQQmKN4tTxZt8CTATQ=='}]
```

### What were you trying to do?
Invoke a `ChatVertexAI` model with `thinking_budget=-1` (dynamic thinking) and `bind_tools()`, then use `response.content` as a string — the standard AIMessage contract.

`langchain-google-vertexai == 3.2.2`

### What did you expect?
`response.content` to be a plain str, as it is when `thinking_budget` is not set.

### What actually happened?
`response.content` is a list containing a single dict: `{'type': 'text', 'text':'...', 'thought_signature': '...'}`.

The root cause is in `_collapse_text_content` (chat_models.py):

#### Current check — requires keys to be exactly {"type", "text"}
```
  if all(
      isinstance(item, dict)
      and item.get("type") == "text"
      and set(item.keys()).issubset({"type", "text"})  # ← fails when thought_signature present
      for item in content
  ):
      return "".join(item.get("text", "") for item in content)
  return content  # ← returns list instead of string
```

When `thinking_budget` is enabled, Gemini attaches a `thought_signature` field to text blocks. The `issubset({"type", "text"})` check fails, so the function returns the raw list instead of collapsing to a string — even when all items are type: text with no separate thinking blocks.

### Suggested fix: 
Allow `thought_signature` as an acceptable key during collapse.

HarishNarava:
Description

When `thinking_budget` is enabled, Gemini attaches a `thought_signature` field to text blocks. The previous `issubset({"type", "text"})` check in `_collapse_text_content` failed when this field was present, causing `response.content` to silently return a list instead of a plain string.

Relevant issues: Fixes #1621

Type: Bug Fix

Changes: 
- Added `thought_signature` as an accepted key in `_collapse_text_content` in `chat_models.py`

Testing:
Added unit tests in `test_chat_models.py` covering:
- Single text block with `thought_signature` collapses to a plain string
- Multiple text blocks with `thought_signature` are joined correctly
- Plain text blocks without `thought_signature` continue to work as before

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/chat_models.py
+++ b/libs/vertexai/langchain_google_vertexai/chat_models.py
@@ -671,7 +671,7 @@ def _collapse_text_content(content: list[Any]) -> str | list[Any]:
     if all(
         isinstance(item, dict)
         and item.get("type") == "text"
-        and set(item.keys()).issubset({"type", "text"})
+        and set(item.keys()).issubset({"type", "text", "thought_signature"})
         for item in content
     ):
         return "".join(item.get("text", "") for item in content)
--- a/libs/vertexai/tests/unit_tests/test_chat_models.py
+++ b/libs/vertexai/tests/unit_tests/test_chat_models.py
@@ -47,6 +47,7 @@ from langchain_google_vertexai._image_utils import ImageBytesLoader
 from langchain_google_vertexai.chat_models import (
     ChatVertexAI,
     _bytes_to_base64,
+    _collapse_text_content,
     _parse_chat_history_gemini,
     _parse_examples,
     _parse_response_candidate,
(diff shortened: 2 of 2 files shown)


mdrxy:
Thanks for digging into this! After tracing the round-trip I think this fix is net-negative and we should close in favor of treating #1621 as working-as-intended.

**Why the list shape exists.** `_parse_response_candidate` (`chat_models.py:718`) deliberately wraps text into a dict when `thought_signature` is present so `_convert_to_prompt` (`chat_models.py:264`) can replay the signed thought on the next turn. That's the whole point of the signature — Gemini requires it back on subsequent turns under `thinking_budget`.

**What this PR does to the round-trip.** Allowing `thought_signature` through `_collapse_text_content` flattens the dict to a plain `str`, which silently strips the signature. Repro:

```
LIST (current main)  ->  parts=1  signatures=[b'SIGN_BLOB']   ✓
STR  (this PR)       ->  parts=1  signatures=[b'']            ✗
```

So the string-content UX in #1621 comes at the cost of breaking signed-thought continuity — the bug just moves from "wrong type" to "silently corrupted multi-turn state."

**How other partners handle this.** Both keep `content` as a `list` whenever reasoning is enabled, because the reasoning state lives on its own block (Anthropic: `{type: "thinking", signature: ...}`, OpenAI Responses: `{type: "reasoning", id: ...}`). The text block stays clean, so collapsing-to-str never has to strip anything. Gemini packs the signature onto the text part itself, so the equivalent of "list when reasoning is on" is "list when `thought_signature` is present" — which is what main already does.

**Suggested resolution.** Close this PR and document in #1621 that under `thinking_budget`, `response.content` is `list[dict]` — matching `ChatAnthropic` with `thinking` and `ChatOpenAI` with `reasoning`. The presence of `thought_signature` is meaningful metadata, not noise to collapse away.

Happy to send a docs PR for the `thinking_budget` page if that's helpful.

PR #1642, opened 2026-03-13
fix(genai): preserve dict field type in nested tool schemas

### LINKED ISSUE #1622: Nest schema field convention bug

JasonCZMeng:
## Bug

Nested `dict` fields in tool schemas are silently converted from `OBJECT` to `STRING` type when the field appears inside a nested model (e.g., `list[Step]` where `Step` has a `dict` field). This causes Gemini to send JSON strings instead of objects, breaking Pydantic validation.

Fixes #1622

## Root Cause

`_get_properties_from_schema` did not preserve the `additionalProperties` flag in its output. When the schema was processed a second time through `_format_json_schema_to_gapic` (via `_dict_to_genai_schema`'s recursive calls), the flag was lost. Without it, the OBJECT→STRING downgrade logic incorrectly triggered for `dict`-typed fields that have no explicit properties.

## Fix

When an OBJECT-typed field has `additionalProperties` set but no explicit `properties`, carry the flag forward in the output dict so subsequent processing passes retain the correct type.

## Test

Added `test_nested_dict_field_preserves_object_type` that verifies `dict` fields nested inside `list[Model]` retain `Type.OBJECT`.

All 28 existing unit tests pass.

### DIFF
--- a/libs/genai/langchain_google_genai/_function_utils.py
+++ b/libs/genai/langchain_google_genai/_function_utils.py
@@ -578,9 +578,7 @@ def _get_properties_from_schema(schema: dict) -> dict[str, Any]:
                 # Preserve additionalProperties flag so that subsequent schema
                 # processing passes don't lose it and incorrectly downgrade
                 # dict-typed fields to STRING.
-                properties_item["additionalProperties"] = v[
-                    "additionalProperties"
-                ]
+                properties_item["additionalProperties"] = v["additionalProperties"]
             else:
                 # Only provide dummy type for object without properties AND
                 # without additionalProperties
--- a/libs/genai/langchain_google_genai/_function_utils.py
+++ b/libs/genai/langchain_google_genai/_function_utils.py
@@ -574,9 +574,16 @@ def _get_properties_from_schema(schema: dict) -> dict[str, Any]:
                     properties_item["required"] = [
                         k for k, v in v_properties.items() if "default" not in v
                     ]
-            elif not v.get("additionalProperties"):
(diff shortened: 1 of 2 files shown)


JasonCZMeng:
Closing this to keep my PR queue focused — no reviewer activity in ~40 days. The change is still mergeable; happy to reopen if a maintainer wants to pick it up, or I'll revisit if the underlying issue resurfaces. Thanks!

PR #1694, opened 2026-04-12
feat(genai): add const/Literal support in genai schema conversion

Serjbory:
## Description

Add support for the JSON Schema `const` keyword in the genai schema conversion pipeline. Pydantic's `Literal['value']` (single-value) generates `{"const": "value"}` in JSON Schema, which was previously silently dropped since `const` is not in the Gemini API's allowed schema fields. This converts `const` to `enum` with a single value, preserving discriminator information for discriminated unions and single-value Literal fields.

## Relevant issues

N/A

## Type

🆕 New Feature

## Changes

- `_format_json_schema_to_gapic`: convert `const` to `enum` during schema formatting, with a guard to not override an existing `enum`
- `_get_properties_from_schema`: convert `const` to `enum` before property processing so it flows through the existing enum handling path

## Testing

- `test_tool_with_literal_single_value_param` — end-to-end: `@tool` with `Literal["greet"]` → OpenAI → GenAI → `tool_to_dict`, verifies `enum == ["greet"]`
- `test_tool_with_pydantic_literal_const` — Pydantic `BaseModel` with `Literal` field via `_convert_pydantic_to_genai_function`
- `test_const_in_anyof_discriminated_union` — `const` inside `anyOf` branches (discriminated union pattern) via `_dict_to_genai_schema`
- `test_const_with_non_string_values` — integer and boolean `const` values via `_format_json_schema_to_gapic`
- `test_const_does_not_override_existing_enum` — when both `const` and `enum` are present, `enum` takes precedence
- `test_const_in_raw_dict_schema` — raw dict schema through `convert_to_genai_function_declarations` full pipeline

All existing tests continue to pass. `make lint` and `make test` pass cleanly.

## Note

- A matching change for the `vertexai` package will be submitted as a separate PR per the contribution guidelines ("PR doesn't touch more than one package").
- Areas requiring careful review: the guard logic ensuring `const` doesn't override a pre-existing `enum` in both `_format_json_schema_to_gapic` and `_get_properties_from_schema`.


Serjbory:
Hello @ninjapenguin , @nfcampos . Please, review this PR, I think that this feature might be pretty valuable for everybody who works with python :)
Also, I included changes from https://github.com/langchain-ai/langchain-google/pull/1588 to pass the CI int. tests, but if it's not ok for you, I can revert them

PR #1708, opened 2026-04-18
fix(genai): wrap batch texts in Content objects for correct embedding count

### LINKED ISSUE #1704: [BUG] GoogleGenerativeAIEmbeddings returns a single vector for batch input in v4.2.2
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
import asyncio, os

# Requires GOOGLE_API_KEY set in environment
async def test():
    e = GoogleGenerativeAIEmbeddings(model="gemini-embedding-2-preview")

    texts = ["hola mundo", "test dos", "test tres", "cuatro", "cinco"]

    # Async batch
    r = await e.aembed_documents(texts)
    print(f"aembed_documents: passed {len(texts)} texts, got {len(r)} vectors")
    # Expected: 5 — Actual: 1

    # Sync batch
    r2 = e.embed_documents(["a", "b", "c"])
    print(f"embed_documents:  passed 3 texts, got {len(r2)} vectors")
    # Expected: 3 — Actual: 1

asyncio.run(test())
```

```shell
N/A (This is a silent failure / data loss bug, no traceback is thrown).
```

### Description

1. What were you trying to do?
Generate embeddings for a batch of text strings using embed_documents and aembed_documents with model gemini-embedding-2-preview (version 4.2.2).

2. What did you expect to happen?
The returned list length should equal the input list length. Passing N strings should return N embedding vectors.

3. What actually happened?
Both embed_documents and aembed_documents always return a list with exactly 1 vector, regardless of batch size. No error is raised — the extra embeddings are silently dropped.

Root cause (identified from source):
In langchain_google_genai/embeddings.py, embed_documents (line ~439) and aembed_documents (line ~557) call:
```
result = self.client.models.embed_content(
    model=self.model,
    contents=batch,   # batch is list[str]
    config=config,
)
```

The Google GenAI SDK interprets contents=list[str] as a single multi-part Content object (treating each string as a Part), resulting in one embedding for the entire batch.

Workaround (confirmed working):
Wrapping each string in a Content object fixes the behavior:

```
from google.genai.types import Content, Part
contents = [Content(parts=[Part(text=t)]) for t in batch]
result = self.client.models.embed_content(model=self.model, contents=contents, config=config)
# Returns len(batch) embeddings correctly
```

gaurav0107:
## Description

When \`GoogleGenerativeAIEmbeddings.embed_documents()\` is called with multiple texts, the Google GenAI SDK receives a bare \`list[str]\`. On the **Vertex AI embedContent path**, the SDK's internal \`t_contents()\` merges all strings into a single multi-part \`Content\`, producing only one embedding for the entire batch regardless of input size.

This fix wraps each text in its own \`Content(parts=[Part(text=t)])\` object before calling the SDK, so it treats them as separate embedding requests.

Additionally, the Vertex AI embedContent path enforces a single-Content-per-request limit (\`ValueError\`). When this is hit, the fix falls back to issuing individual per-text requests, ensuring correct behavior on all three SDK paths:

| Path | Batch behavior |
|------|---------------|
| Gemini Developer API (mldev) | Batch call succeeds — one API call for N texts |
| Vertex AI PREDICT (e.g. \`text-embedding-004\`) | Batch call succeeds |
| Vertex AI embedContent (e.g. \`gemini-embedding-2-preview\`) | Falls back to per-text requests |

Applied to both the sync (\`embed_documents\`) and async (\`aembed_documents\`) code paths.

## Relevant issues

Fixes #1704

## Type

🐛 Bug Fix

## Changes

- Import \`Content\` and \`Part\` from \`google.genai.types\`
- Wrap each batch text in a \`Content\` object before passing to \`embed_content\`
- Add \`ValueError\` fallback for Vertex AI embedContent single-content limit (only catches the specific SDK guard message; re-raises unrelated \`ValueError\`s)
- Wrap per-text fallback calls in the same \`ClientError\`/\`Exception\` handling for consistent \`GoogleGenerativeAIError\` reporting
- Updated unit tests to assert \`Content\` objects are passed instead of bare strings
- Added tests for the Vertex AI fallback path (sync and async)
- Added test verifying unrelated \`ValueError\` propagation

## Testing

- All 16 unit tests pass (\`make test\`)
- \`make lint\` and \`make format\` pass clean (ruff + mypy)
- Tests cover: Content wrapping, Vertex AI fallback, unrelated ValueError propagation, async parity

## Note

> 🤖 Generated with the help of [Claude Code](https://claude.ai/code)

### DIFF
--- a/libs/genai/langchain_google_genai/embeddings.py
+++ b/libs/genai/langchain_google_genai/embeddings.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 import string
@@ -15,6 +16,8 @@ from langchain_google_genai._common import (
     GoogleGenerativeAIError,
     get_user_agent,
 )
+
+logger = logging.getLogger(__name__)
 
 _MAX_TOKENS_PER_BATCH = 20000
 _DEFAULT_BATCH_SIZE = 100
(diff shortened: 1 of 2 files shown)


gaurav0107:
Hi! The integration test `test_context_caching[google_ai]` is failing in CI, but I believe this is unrelated to the changes in this PR (which only touch `embeddings.py` and `test_embeddings.py`).

The failure is in `test_chat_models.py:2521`:
```
assert isinstance(response.content, str)
```

The model (`gemini-3-flash-preview`) now returns a `thought_signature` on text parts, which causes `_parse_response_candidate` to return list-based content blocks instead of a plain string:
```python
[{'type': 'text', 'text': 'The secret number is 747.', 'extras': {'signature': '...'}}]
```

This looks like a pre-existing issue with the test not accounting for Gemini 3+ list-based content format. Could someone help confirm this is not correlated with my embedding changes? Happy to open a separate fix for the test if needed.

PR #1718, opened 2026-04-25
fix(genai): handle Vertex AI single-content limit in batch embeddings

### LINKED ISSUE #1704: [BUG] GoogleGenerativeAIEmbeddings returns a single vector for batch input in v4.2.2

zambradalove:
Resolves #1704

Problem
As reported in the issue, GoogleGenerativeAIEmbeddings.embed_documents() and aembed_documents() were failing silently when processing batches of text. The Google GenAI SDK interprets a bare list[str] as a single multi-part Content object, resulting in exactly one embedding vector regardless of the input size (Silent Data Loss).

Solution
Content Wrapping: Each text string is now explicitly wrapped in its own Content and Part objects before being passed to the SDK. This ensures the SDK treats them as separate requests.

Vertex AI Fallback: Since Vertex AI enforces a strict limit of one Content object per request (for certain models like gemini-embedding-2), the fix includes a ValueError fallback. If the batch request is rejected by Vertex, the code catches the error and automatically issues individual sequential requests.

Quality Assurance
Unit Tests: Updated test_embeddings.py to assert that Content objects are being passed instead of bare strings.

Fallback Tests: Added new sync and async tests specifically for the Vertex AI fallback logic, verifying that individual calls are made after a batch failure.

Error Propagation: Added tests to ensure unrelated ValueErrors are still propagated correctly.

Linting & Formatting: All checks passed via make format, make lint_package, and make lint_tests.

Global Tests: All 260 unit tests passed successfully.

Note: This implementation provides full test coverage for the Vertex AI fallback logic and ensures all CI checks pass locally, providing a robust solution for the silent data loss reported in #1704.

### DIFF
--- a/libs/genai/langchain_google_genai/embeddings.py
+++ b/libs/genai/langchain_google_genai/embeddings.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 import string
@@ -5,7 +6,7 @@ from typing import Any
 
 from google.genai.client import Client
 from google.genai.errors import ClientError
-from google.genai.types import EmbedContentConfig, HttpOptions
+from google.genai.types import Content, EmbedContentConfig, HttpOptions, Part
 from langchain_core.embeddings import Embeddings
 from langchain_core.utils import from_env, secret_from_env
 from pydantic import BaseModel, ConfigDict, Field, SecretStr, model_validator
@@ -15,6 +16,8 @@ from langchain_google_genai._common import (
     GoogleGenerativeAIError,
     get_user_agent,
 )
+
+logger = logging.getLogger(__name__)
 
 _MAX_TOKENS_PER_BATCH = 20000
 _DEFAULT_BATCH_SIZE = 100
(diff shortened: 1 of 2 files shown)


zambradalove:
Hi maintainers, I see the CI failed in test_chat_models.py. Specifically, test_context_caching is failing due to the known issue where Gemini 3 models return a thought_signature in a list format instead of a plain string, and test_streaming_with_multiple_tool_calls seems to be a flaky model behavior.

I want to point out that all 16 integration tests for embeddings passed, confirming that the fix for #1704 is working as expected and doesn't interfere with other modules. This PR is ready for review.

Issue #1752, opened 2026-05-04
feat(genai): support streamFunctionCallArguments / partialArgs for Gemini 3+ streaming


bmd1905:
### Package

- [x] **langchain-google-genai**
- [x] **langchain-google-vertexai** (same code path via `vertexai=True`)

### Checked other resources

- [x] Descriptive title
- [x] Searched docs / API reference
- [x] Searched existing issues — no hits for `streamFunctionCallArguments`, `partialArgs`, `willContinue`, `partial_args`, or "fine-grained tool streaming"
- [x] This is a feature request, not a bug

### Feature Description

Expose Vertex AI / Gemini API's **`streamFunctionCallArguments`** flag and surface the resulting `partialArgs` / `willContinue` chunks as incremental `tool_call_chunks` deltas on `AIMessageChunk`.

Gemini 3 Pro and later support fine-grained streaming of function call arguments, but `langchain-google-genai==4.2.1` (and `main` at the time of writing) does not pass the flag through and therefore always receives a single complete `function_call` Part — i.e., one `tool_call_chunk` containing the full `json.dumps(args)`. Other ecosystem libraries (Vercel AI SDK, LiteLLM) have already added support; LangChain has not.

Google docs:
- [Generate content with the Gemini API in Vertex AI — `tool_config.functionCallingConfig.streamFunctionCallArguments`](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/model-reference/inference)
- [Introduction to function calling — partial args + `willContinue`](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling)

### Use Case

We run multi-turn agents on `vertexai/gemini-3.1-pro-preview` via `ChatGoogleGenerativeAI(streaming=True, vertexai=True)` and stream chunks to a UI through `astream_events`. For Anthropic / OpenAI the UI gets chunk-by-chunk argument streaming (`input_json_delta` / OpenAI tool-call deltas), so users see arguments fill in token-by-token. On Gemini, the entire args object lands in one `tool_call_chunk` — the UI animation is missing and perceived latency is materially worse on long tool calls (e.g. SQL, deep JSON payloads).

The model already produces partials; the SDK already exposes the flag; we just don't have a way to request and translate them in LangChain.

### Proposed Solution

Two-line change at the request site, plus chunk translation:

[...]

PR #1753, opened 2026-05-04
feat(genai): support streamFunctionCallArguments / partialArgs streaming for Gemini 3+

bmd1905:
## Summary

Adds an opt-in `stream_function_call_arguments: bool = False` field to `ChatGoogleGenerativeAI`. When enabled on the streaming path, threads `tool_config.functionCallingConfig.streamFunctionCallArguments=true` into Gemini 3+ requests so the model emits typed `PartialArg` leaf updates (JSONPath + scalar) as it composes a tool call, instead of returning the full `args` dict only at the end.

Closes feature gap raised in #1752. Other ecosystems (Vercel AI SDK, LiteLLM) shipped this; LangChain has not.

## What changes

- **New field** `stream_function_call_arguments: bool = False` on `ChatGoogleGenerativeAI` (defaults off → behavior unchanged for existing users).
- **`_function_utils._tool_choice_to_tool_config`** gains a keyword-only `stream_function_call_arguments` param; only sets the SDK field when `True` (avoids accidental wire-format drift).
- **`_prepare_request` / `_process_tool_config`** gain a `streaming` kwarg. Only `_stream`/`_astream` pass `streaming=True`. Unary `_generate`/`_agenerate` keep the default `False` — Vertex `:generateContent` rejects the flag (see vercel/ai#14314 → vercel/ai#14352), so we must NOT send it on those.
- **`_parse_response_candidate(streaming=True)`** surfaces each `PartialArg` via `additional_kwargs["gemini_partial_args"]` carrying `tool_call_id`, `sdk_call_id` (raw `FunctionCall.id`), `name`, `json_path`, typed `value`, `will_continue`. Standard `tool_call_chunks` continue to carry the SDK-assembled `fc.args` once available — atomic emission, byte-identical to the flag-off path.
- **README** documents the opt-in plus current scope and follow-up.

## Scope (intentionally minimal)

This PR ships the flag wiring and the structured `gemini_partial_args` side channel. **Translating `PartialArg` deltas into incremental `tool_call_chunks`** (the full Anthropic/OpenAI streaming UX where `tool_call_chunks.args` is concat-clean across chunks) is intentionally **out of scope** here. Reason: the Vertex 3.x preview wire format has two sharp edges that need design work to handle faithfully:

1. `FunctionCall.id` is **not populated on the wire** today (verified via raw curl probe against `gemini-3.1-pro-preview`), so chunks cannot be correlated by id.
2. String values arrive **chunked mid-leaf** (e.g. `"$.date"` arrives as `"2026-0"` → `"6-01"` → `""`), requiring per-leaf concatenation rather than replacement.

Building the buffer on assumptions that may shift as the preview stabilizes risks committing to a policy maintainers may want to revisit. Shipping the structured side channel today unblocks downstream consumers (UI previews, observability tooling) that just need the typed deltas; the incremental `tool_call_chunks` translator can land as a follow-up once the format firms up.

## Tests

- **Unit** (`tests/unit_tests/test_streaming_partial_args.py`, 6 tests):
  - Flag-on threads `streamFunctionCallArguments=true` on the stream path
  - Unary path **never** sends the flag (regression guard against vercel/ai#14314)
  - `PartialArg` surfaces via `gemini_partial_args`; no `tool_call_chunks` at that layer
  - Seal chunk on populated `fc.args` produces parseable atomic JSON
  - Flag-off regression keeps single atomic chunk (byte-identical to current)
  - End-to-end `astream` merges chunks to full `args` dict
- **Unit** (`tests/unit_tests/test_function_utils.py`): only sets `stream_function_call_arguments` on `FunctionCallingConfig` when True.
- **Integration** (`@pytest.mark.extended`, Vertex-only): live call to `gemini-3.1-pro-preview` asserts ≥2 `gemini_partial_args` events flow with JSONPath-keyed deltas. Verified green against a real Gemini 3.1 endpoint.

```bash
# Unit
cd libs/genai && make test                                          # 290 → 296 passing
# Integration (requires Vertex creds + Gemini 3 preview access)
TEST_VERTEXAI=only GOOGLE_CLOUD_PROJECT=... \
  uv run --group test --group test_integration pytest --extended \
    -k test_streaming_partial_args_real_call \
    tests/integration_tests/test_chat_models.py
```

## Test plan

- [x] `cd libs/genai && uv run ruff check` clean
- [x] `cd libs/genai && uv run ruff format --check` clean
- [x] `cd libs/genai && make test` — full unit suite green (290 → 296)
- [x] Live integration green against `gemini-3.1-pro-preview` on Vertex (project with Gemini 3 preview allowlist)
- [x] Flag-off path byte-identical to current — regression test included

## Contributing adherence

- [x] Single package (`langchain-google-genai` only — `vertexai` legacy untouched)
- [x] Opt-in default (`stream_function_call_arguments: bool = False`)
- [x] Unit-first tests; integration test gated behind `@pytest.mark.extended`
- [x] No public API removals; field added to existing class

### DIFF
--- a/libs/genai/README.md
+++ b/libs/genai/README.md
@@ -42,6 +42,7 @@ Each entry carries the resolved `tool_call_id`, the raw SDK `FunctionCall.id` (`
 
 #### Caveats
 
+- **Vertex AI backend only.** The Gemini API (`generativelanguage.googleapis.com`) backend rejects this flag at the SDK serializer (`google.genai` raises `ValueError` before any wire call), so it is only useful with `vertexai=True`.
 - The flag is only sent on the streaming endpoint. Non-streaming `invoke` / `ainvoke` calls ignore it because the Vertex `:generateContent` endpoint rejects the option (see [vercel/ai#14314](https://github.com/vercel/ai/issues/14314), [vercel/ai#14352](https://github.com/vercel/ai/pull/14352)).
 - Translating `PartialArg` deltas into incremental `tool_call_chunks` (the full Anthropic-/OpenAI-style streaming experience) is intentionally out of scope for this opt-in: the Vertex 3.x preview wire format omits `FunctionCall.id` and chunks string values mid-leaf, so a faithful translator needs additional design — tracked as a follow-up. The structured side channel above is sufficient for most preview/UI use cases.
 
--- a/libs/genai/README.md
+++ b/libs/genai/README.md
@@ -19,6 +19,32 @@ pip install langchain-google-genai
 
 For full documentation, see the [API reference](https://reference.langchain.com/python/integrations/langchain_google_genai/). For conceptual guides, tutorials, and examples on using these classes, see the [LangChain Docs](https://docs.langchain.com/oss/python/integrations/providers/google#google-generative-ai).
 
+### Streaming tool call arguments (Gemini 3+)
+
+Gemini 3 Pro and later can stream tool/function call arguments as typed leaf updates (`PartialArg`) instead of one final block. Opt in by setting `stream_function_call_arguments=True`; the leaf updates are then surfaced on each `AIMessageChunk` via `additional_kwargs["gemini_partial_args"]`, so consumers can preview values as they arrive without waiting for the full JSON to close.
+
+```python
+from langchain_google_genai import ChatGoogleGenerativeAI
+
+llm = ChatGoogleGenerativeAI(
+    model="gemini-3.1-pro-preview",
+    stream_function_call_arguments=True,
+).bind_tools([book_flight])
+
+async for chunk in llm.astream("Book SFO -> JFK on 2026-06-01."):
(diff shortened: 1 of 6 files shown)


bmd1905:
@mdrxy The `langchain-google-genai-us (llm-integration-tests)` job timed out after 60m on commit `11f9282`. The diff doesn't touch any default code path (`stream_function_call_arguments` defaults to `False`, behavior is byte-identical to `main` when off), and the one new integration test (`test_streaming_partial_args_real_call`) now skips on the default `google_ai` backend, so the timeout looks unrelated to this PR.

Could you re-run the job? Happy to dig in if it stalls again on a fresh run.

mdrxy:
Thanks for opening this.

That said, this reads as AI-generated and submitted without a human in the loop, which makes it hard for us to engage productively. We'd rather have a shorter issue you've read through and can speak to directly than a thorough one you haven't.

Closing for now.

PR #1766, opened 2026-05-11
chore(genai): Update google-genai dependency to support 2.x

sararob:
## Description

Update the `google-genai` dependency to support version 2.x.

[The 2.0 release](https://github.com/googleapis/python-genai/releases/tag/v2.0.0) of the `google-genai` SDK introduces breaking changes that are strictly isolated to the Interactions API. Core functionality and standard usage (like this langchain integration) will remain backwards compatible with this version jump. We plan to update the `google-genai` dep in the [google-cloud-aiplatform SDK](https://github.com/googleapis/python-aiplatform/) to include 2.x, but wanted to make sure this langchain integration was updated first so folks using both would be unaffected.

## Type

🚄 Infrastructure


sararob:
@mdrxy could you take a look at this when you get a chance? We'd like to upgrade the version of `google-genai` used here to >= 2.x. I wasn't exactly sure how to fix the failing tests, feel free to advise or take over the PR if that's easier.

For more context, see the langchain tests failing in https://github.com/googleapis/python-aiplatform/pull/6776. Thanks!

libs/genai/pyproject.toml
```diff
@@ -13,7 +13,7 @@ version = "4.2.2"
 requires-python = ">=3.10.0,<4.0.0"
 dependencies = [
     "langchain-core>=1.3.2,<2.0.0",
-    "google-genai>=1.65.0,<2.0.0",
+    "google-genai>=1.65.0,<3.0.0",
```
lkuligin, line 16: should it be >=2.0.0 maybe?

sararob: I think we should keep it at `>=1.65.0` for now since the Agent Platform SDK (`google-cloud-aiplatform`) still [allows](https://github.com/googleapis/python-aiplatform/blob/main/setup.py#L329-L330) versions < 2.0. What do you think?

lkuligin (approved):

Issue #1692, opened 2026-04-10
Remote HTTP/HTTPS image_url handling can block request preparation indefinitely


zhaoce:
### Package (Required)

- [x] langchain-google-genai
- [ ] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
from langchain_core.messages import HumanMessage
from langchain_google_genai import ChatGoogleGenerativeAI

llm = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash",
    google_api_key="test",
    timeout=5,
)

message = HumanMessage(
    content=[
        {"type": "text", "text": "Describe this image."},
        {
            "type": "image_url",
            "image_url": {
                # If the upstream server behind this URL hangs, LangChain will block while
                # fetching it, and that can also hang our own service request.
                "url": "https://example.com/test.png"
            },
        },
    ]
)

llm.invoke([message])
```

### Error Message and Stack Trace (if applicable)

```shell
There is usually no Python exception in the hanging case. The call blocks during request preparation while fetching the remote HTTP/HTTPS image_url.

A representative stack looks like this:
_bytes_from_url (langchain_google_genai/_image_utils.py)
load_part (langchain_google_genai/_image_utils.py)
```

### Description

When an HTTP/HTTPS `image_url` is included in the message content, `ChatGoogleGenerativeAI` fetches it during request preparation using blocking `requests.get(...)`.

This fetch does not pass a timeout, so a slow or stalled remote URL can block the call indefinitely before the Gemini API request is even sent. `ChatGoogleGenerativeAI(timeout=...)` does not seem to protect this path.

PR #1826, opened 2026-06-04
fix(vertexai): support timeout in `ImageBytesLoader` URL fetches

HumphreySun98:
## Description

`ImageBytesLoader._bytes_from_url` in `libs/vertexai/langchain_google_vertexai/_image_utils.py` calls `requests.get(url)` without a `timeout=`, so a slow or stalled remote image/PDF URL can block request preparation indefinitely before any Vertex API call is issued — the same class of bug fixed on the genai side in #1693.

This PR closes that gap **end-to-end** for the public chat models, matching #1693:

1. **Loader** — `ImageBytesLoader` gains a keyword-only `timeout: float | None = None` that is threaded into `requests.get`.
2. **Gemini (`ChatVertexAI`)** — `_image_bytes_loader_client` now builds `ImageBytesLoader(project=self.project, timeout=self.timeout)`, so every media fetch on the standard chat path respects the model's configured `timeout`.
3. **Claude on Vertex (`ChatAnthropicVertex`)** — `timeout` is threaded through `_format_messages_anthropic` → `_format_message_anthropic` → `_format_image` / `_format_image_content_block`, and `_format_params` passes `self.timeout`. Covers both `image` and `image_url` blocks.
4. **`create_context_cache`** — reuses `model._image_bytes_loader_client` instead of constructing a fresh loader, so it inherits the timeout automatically.

Vision (image-generation) models are intentionally out of scope: those classes expose no `timeout` field, and the failure mode here is chat media fetches.

## Relevant issues

Parallel to the genai analog #1692 / #1693 (same code path, same missing `timeout`). No vertexai-specific issue was open; this closes the gap so both packages share the API.

## Type

🐛 Bug Fix

## Changes

- `_image_utils.py`: keyword-only `timeout` on `ImageBytesLoader`, threaded into `requests.get`.
- `chat_models.py`: `_image_bytes_loader_client` passes `self.timeout`.
- `_anthropic_utils.py`: `timeout` threaded through the message-formatting chain.
- `model_garden.py`: `ChatAnthropicVertex._format_params` passes `self.timeout`.
- `utils.py`: `create_context_cache` reuses the model's loader.

## Testing

Behavioral unit tests drive the full chain with `requests.get` mocked and assert the timeout reaches it (and that the default stays unbounded / backwards compatible):

- `tests/unit_tests/test_image_utils.py`: loader forwards / omits `timeout`.
- `tests/unit_tests/test_anthropic_utils.py`: `_format_messages_anthropic` threads the timeout to the remote fetch (`test_format_messages_anthropic_threads_media_fetch_timeout` + default sibling).
- `tests/unit_tests/test_chat_models.py`: `ChatVertexAI(timeout=...)` wires it into the loader (`test_image_bytes_loader_client_inherits_model_timeout`).

```
$ uv run pytest libs/vertexai/tests/unit_tests/test_image_utils.py libs/vertexai/tests/unit_tests/test_anthropic_utils.py -q
... passed
```

`ruff check` and `ruff format --check` pass on all changed files.

## Note

`timeout` is keyword-only per the `CLAUDE.md` guidance for new public-API parameters; existing call sites are unchanged. No dependencies added.

---

Disclaimer: this PR was prepared with the assistance of an AI agent (Claude Code). All code and test changes were reviewed by the author before submission.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
@@ -78,7 +78,9 @@ def _create_usage_metadata(anthropic_usage: BaseModel) -> UsageMetadata:
     )
 
 
-def _format_image(image_url: str, project: str | None) -> dict:
+def _format_image(
+    image_url: str, project: str | None, timeout: float | None = None
+) -> dict:
     """Formats a message image to a dict for Anthropic API."""
     regex = r"^data:(?P<media_type>(?:image|application)/.+);base64,(?P<data>.+)$"
     match = re.match(regex, image_url)
@@ -89,7 +91,7 @@ def _format_image(image_url: str, project: str | None) -> dict:
             "data": match.group("data"),
         }
     if validators.url(image_url):
-        loader = ImageBytesLoader(project=project)
+        loader = ImageBytesLoader(project=project, timeout=timeout)
         image_bytes = loader.load_bytes(image_url)
         path = urllib.parse.urlparse(image_url).path
         raw_mime_type = path.split(".")[-1].lower()
@@ -106,7 +108,7 @@ def _format_image(image_url: str, project: str | None) -> dict:
         }
     if image_url.startswith("gs://"):
         # Gets image and encodes to base64.
-        loader = ImageBytesLoader(project=project)
+        loader = ImageBytesLoader(project=project, timeout=timeout)
         part = loader.load_part(image_url)
         if part.file_data.mime_type:
             mime_type = part.file_data.mime_type
@@ -143,7 +145,9 @@ def _format_text_content(text: str) -> dict[str, str | dict[str, Any]]:
 
 
 def _format_message_anthropic(
-    message: HumanMessage | AIMessage | SystemMessage, project: str | None
+    message: HumanMessage | AIMessage | SystemMessage,
+    project: str | None,
+    timeout: float | None = None,
 ):
     """Format a message for Anthropic API.
 
@@ -187,7 +191,7 @@ def _format_message_anthropic(
                         new_block[copy_attr] = block[copy_attr]
 
                 if block["type"] == "image":
-                    content.append(_format_image_content_block(block, project))
+                    content.append(_format_image_content_block(block, project, timeout))
                     continue
 
                 if block["type"] == "text":
@@ -236,7 +240,7 @@ def _format_message_anthropic(
 
                 if block["type"] == "image_url":
                     # convert format
-                    source = _format_image(block["image_url"]["url"], project)
+                    source = _format_image(block["image_url"]["url"], project, timeout)
                     if source["media_type"] == "application/pdf":
                         doc_type = "document"
                     else:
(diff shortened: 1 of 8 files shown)


chuenchen309:
Thanks for tightening this up — the change itself looks correct and backwards compatible (keyword-only param, default path unchanged), and the tests cover both branches. A couple of observations, happy to be wrong on either:

1. **The timeout doesn't seem to reach users of the public models.** As far as I can tell, every internal construction of `ImageBytesLoader` still passes only `project=`:

   - `chat_models.py:1982`, `vision_models.py:60` / `386`, `utils.py:58`, `_anthropic_utils.py:92` / `109`

   The URL fetch this guards (`_bytes_from_url` via `load_bytes`) is reached from `_anthropic_utils._format_image` and the vision models, but none of them forward a timeout. Since `ChatVertexAI` already has a `timeout` field, it might be worth threading it through at those construction sites (e.g. `ImageBytesLoader(project=self.project, timeout=self.timeout)`, and passing it through `_format_image`) so the "stalled media URL blocks indefinitely" case is actually fixed end-to-end — otherwise the new arg is only reachable by constructing the loader directly. If a follow-up PR is the plan, totally reasonable — just flagging it.

2. Minor: `requests.get(url, timeout=None)` behaves identically to omitting the kwarg, so the `if self.timeout is None` split in `_bytes_from_url` is functionally redundant — a single `requests.get(url, timeout=self.timeout)` would do (the branch is really just there to satisfy the exact-call assertion in the default test). Purely cosmetic, feel free to ignore.

Neither blocks; the core fix looks right.

HumphreySun98:
Thanks @chuenchen309, both are fair — especially the first one. You're right that the loader arg was unreachable from the public models, which made the fix cosmetic. I've pushed a follow-up commit that threads the timeout end-to-end, so the "stalled media URL blocks the turn" case is actually closed:

- **Gemini (`ChatVertexAI`)** — `_image_bytes_loader_client` now constructs `ImageBytesLoader(project=self.project, timeout=self.timeout)`, so all three call sites (`chat_models.py:2227/2445/2508`) inherit the model's timeout.
- **Claude on Vertex (`ChatAnthropicVertex`)** — added a keyword `timeout` to `_format_messages_anthropic` → `_format_message_anthropic` → `_format_image` / `_format_image_content_block`, and `_format_params` now passes `self.timeout`. This covers both the `image` and `image_url` block paths through `_anthropic_utils`.
- **`create_context_cache`** — now reuses `model._image_bytes_loader_client` instead of building a fresh loader, so it picks up the timeout for free.
- **Vision models** — intentionally left alone; those classes have no `timeout` field, and the concern here is chat media fetches.

New behavioral tests drive the whole chain (mock `requests.get`, assert the timeout reaches it): `test_format_messages_anthropic_threads_media_fetch_timeout` (+ its default-unbounded sibling) and `test_image_bytes_loader_client_inherits_model_timeout`. This now mirrors what genai #1693 does for its `_convert_to_parts` chain.

On the second (cosmetic) point — you're technically correct that `requests.get(url, timeout=None)` is identical to omitting the kwarg. I kept the `if self.timeout is None` split purely to mirror #1693's loader byte-for-byte so the two packages stay in lock-step; happy to collapse it in both PRs together if you'd rather have the simpler form.

Disclaimer: this follow-up, like the original PR, was prepared with the assistance of an AI agent (Claude Code); all changes were reviewed by me before pushing.

PR #1368, opened 2025-11-19
chore(genai,vertexai,community): Making integration tests async #960

RishikaGhosh:
## Description
Converts all synchronous integration tests to asynchronous tests across the langchain-google repository to enable parallel test execution with pytest-xdist.

Fixes #960

### Infrastructure
- Added `pytest-xdist>=3.8.0,<4.0.0` to test dependencies in genai and community packages
- Updated Makefiles to use `-n auto` flag for parallel test execution

#### Genai Package (~20+ tests)
- [test_chat_models.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_chat_models.py:0:0-0:0): All chat model tests
- [test_llms.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_llms.py:0:0-0:0): All LLM tests
- [test_embeddings.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_embeddings.py:0:0-0:0): All embedding tests
- [test_callbacks.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_callbacks.py:0:0-0:0): Streaming callback tests
- `test_tools.py`: Tool tests
- `test_function_call.py`: Function call tests

#### Vertexai Package (~30+ tests)
- [test_chat_models.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_chat_models.py:0:0-0:0): 37+ chat model tests
- [test_llms.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_llms.py:0:0-0:0): All LLM tests
- [test_embeddings.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_embeddings.py:0:0-0:0): All embedding tests
- [test_callbacks.py](cci:7://file:///Users/rishikaghosh/Documents/Open%20Source/langchain-google/libs/vertexai/tests/integration_tests/test_callbacks.py:0:0-0:0): All callback tests
[...]

PR #1840, opened 2026-06-14
chore: migrate integration tests to async

### LINKED ISSUE #960: Improvements: Integration tests should be async
The integration tests currently take a very long time to execute. However they are all I/O bound i.e. calling APIs and waiting for responses. This is very inefficient when running tests in series.

We should expect to see significant speed ups by making the integration tests async and using the `x-dist` plug-in for `pytest` ([link](https://pypi.org/project/pytest-xdist/)).

Shaan-alpha:
Resolves #960. Migrated all integration tests in genai, ertexai, and community packages to be asynchronous. This improves testing efficiency. pytest-asyncio handles the async execution.


Shaan-alpha:
Hello @baskaryan. I have resolved the integration test failures by properly migrating the blocking calls to their async equivalents using an AST transformation. The tests now pass and the parallel execution via xdist is functional. Please let me know if any further adjustments are required. Thank you for your time.

Shaan-alpha:
Hi @baskaryan, all three integration suites (community / genai / vertexai) are green now, and the merge conflict with `main` is resolved. Summary of what was fixed:

- **community** — `test_psc_feature_store` called `await online_store.adelete()`, but `online_store` is a Vertex AI SDK `FeatureOnlineStore` (no async API), so it now uses the sync `delete()`. I also broadened the Sheets tests' existing skip guard to catch the CI key's `API_KEY_SERVICE_BLOCKED` error (it previously matched only `SERVICE_DISABLED`), so they skip as intended rather than failing.
- **genai** — `test_model_methods_without_eventloop` drives the async methods via `asyncio.run()`, so it must stay synchronous; making it `async def` caused `asyncio.run() cannot be called from a running event loop`. Reverted it to `def` and restored the sync `invoke()` branch.
- **conflicts** — merged latest `main` and resolved the two vertexai test conflicts by combining the async conversion (`await ainvoke`) with `main`'s new content-block assertions / single-arg check helpers.

The occasional 503s on the image/search tests are absorbed by the existing retries. Ready for another look whenever you have a chance — thanks!

Shaan-alpha:
Resolved the new conflicts with `main` (genai `test_chat_models`, vertexai `test_chat_models` and `test_llms`). I kept the async conversion (`await ainvoke` / `agenerate`) while adopting main's updates in each case:

- genai code-execution test: main's simplified `assert isinstance(response, AIMessage)` follow-up assertion.
- vertexai `test_candidates` / `test_vertex_generate_multiple_candidates`: main's `_DEFAULT_MODEL_NAME` in place of the deprecated `text-bison@001`.

`make test`, `make lint`, and the compile jobs are green across 3.10–3.14 for all three packages; ruff is clean on the resolved files.

anooprh:
@Shaan-alpha There were two concerns raised with #960 which I think are valid here as well. Just summarized it in that PR (at https://github.com/langchain-ai/langchain-google/pull/1368#issuecomment-4765348362). But basically this gist is

i. some sync tests are still needed. 100% of the tests cannot be moved to async
ii. `-n auto` may cause rate limiting. So limit it to either `-n 4` or `-n 8`

Shaan-alpha:
Thank you, @anooprh — both points are well taken, and I have addressed them in `cfc3d12`.

### 1. Sync path coverage

You are right that converting `invoke` → `ainvoke` across the board leaves the sync transport untested, which would hide a sync-only regression (e.g. a timeout or header applied only to `self.client`). Rather than revert the migration, I have made the key APIs exercise **both** paths:

- **vertexai `invoke`** — `test_vertexai_single_call` is now parametrized over `is_async=[False, True]`, so it runs both `model.invoke()` and `model.ainvoke()`.

For reference, a few other canonical paths already cover both directions and were left intact:

- **genai `invoke` / `batch`** — `test_chat_google_genai_invoke` and `test_chat_google_genai_batch` are parametrized on `is_async`.
- **vertexai `generate`** — `test_vertexai_generate` calls both `generate` and `agenerate`.
- **vertexai `stream`** — `test_vertexai_stream` (sync) sits alongside `test_vertexai_astream` (async).

So across the suite the sync transport stays covered for invoke, batch, generate, and stream. Happy to extend this to any further API you consider critical.

### 2. Rate limiting

I bounded the only unbounded parallelism: genai's `integration_tests` target now uses `-n 4` instead of `-n auto`. The `--retries 3 --retry-delay 1` flags you suggested are already present there, and the community/vertexai targets run serially (no `-n`), so there is no remaining unbounded fan-out. The vertexai suite additionally throttles through the existing `InMemoryRateLimiter(requests_per_second=1.0)`.

I chose `-n 4` as the more conservative of the two values you suggested — glad to switch to `-n 8` if you would prefer the extra throughput.

Finally, regarding the `aevaluate()` note from #1368: that change is not part of this PR — the diff here is limited to the test files plus the xdist test dependency — so there is nothing to remove on this branch.

Thanks again for the detailed and helpful review.

Shaan-alpha:
@anooprh gentle bump on this one — anything else you'd like me to address before a re-review? Happy to. Thanks!

anooprh:
hey @Shaan-alpha , I'm not a maintainer/approver for this repo. 

I was just trying to contribute to the issue and saw that since you had already created PR, I left my comments here.

Shaan-alpha:
Hi @mdrxy, could you take a look at this when convenient? It migrates the integration tests to async; the earlier suggestions (keeping key sync-path coverage and capping xdist workers) were addressed back in June, and all 35 checks are passing. It has been quiet for a few weeks and I would love to know whether anything else is needed to move it forward. Thanks!

Shaan-alpha:
Hey @ccurme — mind taking a look at this when you get a chance?

Rebased on main and cleared the Copilot pass. The useful bit from it: ten tests here were `async def` but never awaited anything, since `get_documents`, `add_texts_with_embeddings` and `batch_search` have no async counterparts. Those are back to sync, which also lines up with the earlier point that not everything can move to async.

Worth flagging on CI: `community-us` is red, but it was already red on main before this branch — #1927 (cryptography 50.0.0) and #1926 (aiohttp 3.14.3) both got merged with that job failing. If any of the other integration jobs are genuinely broken by my changes I'd love to fix them, but I can't see the Cloud Build logs from outside the org, so a paste would help a lot.

Also noticed #1943 is doing the same migration for #960. Happy to defer to it or fold this one in — whichever is less work on your side.

Shaan-alpha:
Merged main in, the branch had conflicted.

The only collision was in `libs/genai/tests/integration_tests/test_chat_models.py`, where `test_agentic_video_understanding_youtube` landed directly above `test_chat_google_genai_single_call_with_history`, which this branch had converted to async. Kept the new test exactly as written and restored the async signature on the one below it, whose body was already using `ainvoke`.

I left `test_agentic_video_understanding_youtube` synchronous rather than migrating it, since it arrived after this PR was opened and it is your test to decide on. Happy to convert it too if you would rather the file be uniform.

Ruff check and format are clean, 52 async tests unchanged, and the lockfiles merged without damage. Mergeable again.

@mdrxy when you get a chance.

PR #1868, opened 2026-06-22
fix(community): resolve document_ids MIME types via files().get (#1867)

### LINKED ISSUE #1867: GoogleDriveLoader.load() fails with HTTP 400 for document_ids (invalid files.list query)
## What happens
Loading files by `document_ids` raises an HTTP 400 "Invalid Value". The call fails for every id, so no document loads.

## Why
`_load_documents_from_ids()` builds a `files.list` query that filters on `id`:

```python
q = " or ".join(f"id = '{doc_id}'" for doc_id in self.document_ids)
results = (
    service.files()
    .list(
        q=q,
        fields="files(id, mimeType)",
        includeItemsFromAllDrives=True,
        supportsAllDrives=True,
    )
    .execute()
)
```

The Drive API `q` parameter has no `id` field, so the request is rejected with 400. To fetch a file by its id you must use `files.get`, not `files.list`. The error happens at `.execute()`, before the per-MIME dispatch below it runs — so the Sheets/PDF handling from #1644 is never reached.

## Where
`libs/community/langchain_google_community/drive.py` → `_load_documents_from_ids()` (added in #1644).

## Affected versions
4.0.0, 5.0.0, and `main` (every release since #1644). 3.x and earlier are not affected.

I confirmed this by inspecting the released wheels:
- **3.0.5** — no `files.list`, uses a simple per-id loop → works
- **4.0.0 / 5.0.0** — build the `id`-based `files.list` query → fail

## How to reproduce
```python
loader = GoogleDriveLoader(document_ids=["<real_doc_id>"], credentials=creds)
loader.load()  # HttpError 400 "Invalid Value"
```
Note: this only fails against the real Drive API. Tests that mock the Drive service do not catch it.

## Suggested fix
Resolve each MIME type with `files.get(fileId=doc_id, fields="mimeType")` per id, instead of the `files.list` query. This keeps the per-type dispatch from #1644 but uses a valid call.

lcmartinezdev:
## Description
`GoogleDriveLoader.load()` fails with HTTP 400 for any `document_ids`. `_load_documents_from_ids` resolves MIME types with `files().list(q="id = '...'")`, but the Drive API `q` parameter has no `id` field, so the request is rejected with "Invalid Value". Loading by `document_ids` is broken against the real API on every release since #1644 (4.0.0, 5.0.0, and `main`). 3.x is not affected. Mocked unit tests did not catch it because the Drive service is stubbed.

## Fix
Resolve each id with `files().get(fileId=..., fields="mimeType")` — the correct way to fetch a file by id. The per-type dispatch added in #1644 (Sheets → `_load_sheet_from_id`, PDF → `_load_file_from_id`, else → `_load_document_from_id`) is unchanged.

## Tests
- Updated the mock service to stub `files().get()` (the call the loader now makes).
- Added a regression test asserting MIME resolution uses `files().get(fileId=...)` and never `files().list`.
- All unit tests pass; `ruff` and `mypy` are clean.

## Note
This makes one `files().get` call per id instead of a single `files().list`. That is required, since `files().list` cannot query by id, and `document_ids` lists are typically small.

Fixes #1867

### DIFF
--- a/libs/community/langchain_google_community/drive.py
+++ b/libs/community/langchain_google_community/drive.py
@@ -542,24 +542,17 @@ class GoogleDriveLoader(BaseLoader, BaseModel):
         creds = self._load_credentials()
         service = build("drive", "v3", credentials=creds)
 
-        q = " or ".join(f"id = '{doc_id}'" for doc_id in self.document_ids)
-        results = (
-            service.files()
-            .list(
-                q=q,
-                fields="files(id, mimeType)",
-                includeItemsFromAllDrives=True,
-                supportsAllDrives=True,
-            )
-            .execute()
-        )
-        mime_types: Dict[str, str] = {
-            f["id"]: f["mimeType"] for f in results.get("files", [])
-        }
-
         documents: List[Document] = []
         for doc_id in self.document_ids:
-            mime_type = mime_types.get(doc_id)
+            # Resolve the MIME type with files().get. The Drive API `q` parameter has no
+            # `id` field, so a files().list(q="id = '...'") lookup is rejected with HTTP
+            # 400; files().get is the correct way to fetch a file by its id.
+            file = (
+                service.files()
+                .get(fileId=doc_id, fields="mimeType", supportsAllDrives=True)
+                .execute()
+            )
+            mime_type = file.get("mimeType")
             if mime_type == "application/vnd.google-apps.spreadsheet":
                 documents.extend(self._load_sheet_from_id(doc_id))
             elif mime_type == "application/pdf" or self.file_loader_cls is not None:
--- a/libs/community/tests/unit_tests/test_drive.py
+++ b/libs/community/tests/unit_tests/test_drive.py
@@ -43,10 +43,11 @@ def test_drive_invalid_scope() -> None:
 
 
 def _make_mock_service(mime_type: str, file_id: str) -> MagicMock:
-    """Return a mock Drive service whose files().list() returns one file entry."""
+    """Return a mock Drive service whose files().get() returns one file's mimeType."""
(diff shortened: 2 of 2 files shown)


lcmartinezdev:
### Why `files.get` per id (and not a single request)

Per the Drive API docs, `id` is not a valid `files.list` query term. The supported `q` terms are:

> `name, fullText, mimeType, modifiedTime, viewedByMeTime, trashed, starred, parents, owners, writers, readers, sharedWithMe, createdTime, properties, appProperties, visibility, shortcutDetails.targetId`

So there is no "fixed query" version of `files.list(q="id = '...'")` — you fetch a file by id with `files.get`, not with search. That leaves two shapes:

1. **`files.get` per id** (this fix) — N calls, simple and standard.
2. **Batch HTTP request** — one round-trip for up to 100 ids, but more error-handling code, and it does **not** reduce quota: the docs state *"n requests batched together count as n requests, not one."*

Batching would only save round-trip latency, not API usage, so it adds complexity for a benefit most callers don't need (id lists are usually short). I went with `files.get` per id: it is the documented way to fetch by id and keeps the change minimal. A batched variant can be added later if a caller needs to load many ids at once.

Docs: [file-specific query terms](https://developers.google.com/workspace/drive/api/guides/ref-search-terms), [batching](https://developers.google.com/workspace/drive/api/guides/performance).

lcmartinezdev:
The failing `community-us` check is unrelated to this PR: two Sheets integration tests fail with `API_KEY_SERVICE_BLOCKED` (403) on the Sheets API, while this PR only touches the Drive loader. Could a maintainer take a look? Thanks!

PR #1878, opened 2026-07-05
fix(genai): propagate model-level `media_resolution` to the request

jawwad-ali:
`media_resolution` set on the `ChatGoogleGenerativeAI` constructor is silently dropped from the request. `_build_base_generation_config` builds every other generation parameter from the instance (`temperature`, `seed`, `response_modalities`, …) but never reads `self.media_resolution`, so a model-level value never reaches the API. Only the per-call keyword works today, and only incidentally — it survives as an unconsumed kwarg splatted onto `GenerateContentConfig`.

Model-level support originally shipped in #1259; the later refactor of `_prepare_params` dropped the line that read `self.media_resolution`. The field is still a public constructor parameter and still appears in `_identifying_params`.

The fix adds `media_resolution` to the base config alongside `seed`. When unset it is filtered out by the existing `if v is not None` comprehension, and because it now enters `params.model_fields_set` it is excluded from the leftover-kwarg pass — so the per-call keyword still works and still takes precedence.

**Careful review:** the per-call path — `media_resolution` now flows through `params` rather than `remaining_kwargs`, so it is not double-applied; `test_call_level_media_resolution_overrides_model_level` locks the precedence.

### Release note

`media_resolution` set on the `ChatGoogleGenerativeAI` constructor is now sent to the model instead of being silently ignored. Per-call `media_resolution` continues to work and takes precedence.

### Tests

No-network unit tests in `test_chat_models.py`: the model-level value reaches the request, a per-call value overrides it, and it is omitted when unset.

---

*Disclosure: this change was prepared with the help of an AI agent (Claude Code); the fix, tests, and local verification (`ruff format` / `ruff check` / `mypy` / unit tests) were reviewed before submission.*

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -2820,6 +2820,7 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
                 "response_modalities", self.response_modalities
             ),
             "seed": kwargs.get("seed", self.seed),
+            "media_resolution": kwargs.get("media_resolution", self.media_resolution),
         }
 
         # Convert response modalities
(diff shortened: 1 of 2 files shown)


antoremin:
+. hit the same issue.

PR #1882, opened 2026-07-07
fix(genai): remove default temperature for gemini models

markmcd:
Newer models will have a fixed temperature and start throwing errors when temperatures are specified, so we no longer want to specify a default.

All gemini-3 series models use a temperature of 1.0 by default, so that does not need to be specified.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -2505,20 +2505,10 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
 
     @model_validator(mode="after")
     def validate_environment(self) -> Self:
-        """Validates params and builds client.
-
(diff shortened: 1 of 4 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -2505,21 +2505,11 @@ def _set_langchain_google_genai_version(self) -> Self:
...
-        """
+        """Validates params and builds client."""
         if self.temperature is not None and not 0 <= self.temperature <= 2.0:
             msg = "temperature must be in the range [0.0, 2.0]"
             raise ValueError(msg)
 
-        if "temperature" not in self.model_fields_set and _is_gemini_3_or_later(
-            self.model
-        ):
-            self.temperature = 1.0
```
ccurme, line 2521: if 1.0 is the effective default, should we just change this to `None`, to avoid behavioral changes for gemini 2.5 models?

ccurme (approved):

lkuligin (approved):

Issue #1256, opened 2025-10-10
400 Bad Request when using Claude Sonnet on GCP


acurat:
Using Claude Sonnet 4 on GCP 

And using `ChatAnthropicVertex` from package `langchain_google_vertexai`

Immediately after the first subagent responds, the graph fails with following the error

Anyone know how to fix it?

`anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.6.content.0.tool_result.content.0.text.index: Extra inputs are not permitted'}, 'request_id': 'req_vrtx_011CTzH5cU85qEobDpNtgbp7'}`

nhuang-lc:
Hey, can you send the exact code snippet you use to initialize your model with ChatAnthropicVertex so I can reproduce?

yann7962:
Hello

I have the same issue. 

The code use to initialize the model is :

claude_sonnet = init_chat_model(
    "google_anthropic_vertex:claude-sonnet-4@20250514", 
    project="############",
    max_tokens=4096,
    location="us-east5",
    temperature=0.28,
    max_retries=3,
    streaming=True
)

used in

agent = create_deep_agent(
    tools=[get_available_assets, get_asset_samples, get_asset_customizations, validate_terraform_config],
    instructions=main_instructions,
    model=claude_sonnet,  # Use the custom model for main agent
    subagents=[
        asset_collection_agent,
        global_config_agent, 
        sample_selection_agent,
        customization_agent,
        terraform_generation_agent
    ],
).with_config({"recursion_limit": 1000})

I'm currently using a workaround to bypass this issue. 

On the _astream method of the langchain_google_vertexai -> model_garden.py -> ChatAnthropicVertex Class 

        # @retry_decorator
        # async def _astream_with_retry(**params: Any) -> Any:
        #     params.pop("stream", None)
        #     return await self.async_client.messages.create(**params, stream=True)

        ########### Fix ToolMessage content formatting ###########
        
        def _fix_tool_message_content(messages: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
            """Fix ToolMessage content to ensure compatibility with Anthropic API."""
            fixed_messages = []
            
            for message in messages:
                if message.get("role") == "user" and "content" in message:
                    content = message["content"]
                    if isinstance(content, list):
                        fixed_content = []
                        for item in content:
                            if item.get("type") == "tool_result":
[...]

PR #1886, opened 2026-07-11
fix(vertexai): download url images to base64 for Claude on Vertex

### LINKED ISSUE #1088: Claude in Vertex AI does not support URL-based image ingestion
### Description
Passing an image as an URL works with Anthropic's API and with Gemini (via Vertex) too. 
But when using Claude via Vertex AI we get an error. (It works with base64)
Google's Python ADK seems to have the same problem : https://github.com/google/adk-python/issues/2466
It has already been noted and fixed by vertex AI : https://github.com/vercel/ai/pull/6618

### Steps to Reproduce
```python

image_url = "https://upload.wikimedia.org/wikipedia/commons/thumb/2/2e/Morenas_negras_%28Muraena_augusti%29%2C_franja_marina_Teno-Rasca%2C_Tenerife%2C_Espa%C3%B1a%2C_2022-01-09%2C_DD_52.jpg/1000px-Morenas_negras_%28Muraena_augusti%29%2C_franja_marina_Teno-Rasca%2C_Tenerife%2C_Espa%C3%B1a%2C_2022-01-09%2C_DD_52.jpg"


def multiply(a: int, b: int) -> int:
    """Multiply two numbers"""
    return a * b


agent = create_react_agent(
    model=agent_model,
    # pass the tool that can update state
    tools=[multiply],
)

print(
    agent.invoke(
        {
            "messages": [
                {
                    "role": "user",
                    "content": [
                        {"type": "text", "text": "Describe this image"},
                        {
                            "type": "image",
                            "source_type": "url",
                            "url": image_url,
                        },
                    ],
                }
            ]
        }
    )
)
``` 

### Expected Behavior

Claude on Vertex should accept URL-based images, just like Anthropic's API and Gemini on Vertex

### Actual Behavior

Error returned:
```shell
Retrying langchain_google_vertexai.model_garden.ChatAnthropicVertex._generate.<locals>._completion_with_retry_inner in 4.0 seconds as it raised BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.0.content.1.image.source.base64.data: URL sources are not supported'}}.
Retrying langchain_google_vertexai.model_garden.ChatAnthropicVertex._generate.<locals>._completion_with_retry_inner in 4.0 seconds as it raised BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': 'messages.0.content.1.image.source.base64.data: URL sources are not supported'}}.
``` 

Environment

langchain version: 0.3.27

Python version: 3.13

Vertex AI model: claude-sonnet-4@20250514

chuenchen309:
## Why

Claude on Vertex AI rejects url-type image sources (unlike Anthropic's direct API and Gemini-via-Vertex), so passing an image by http(s) URL to `ChatAnthropicVertex` fails while base64 works. Reported in #1088; the same Vertex limitation is noted for Google's ADK ([google/adk-python#2466](https://github.com/google/adk-python/issues/2466)) and was fixed the same way in the Vercel AI SDK ([vercel/ai#6618](https://github.com/vercel/ai/pull/6618)).

`_format_image_content_block` passed non-`data:` URLs straight through as a `{"type": "url"}` source. This routes them through `_format_image` instead, which already downloads http(s) URLs (and handles `data:` and `gs://`) and inlines them as base64 — the same treatment `data:` URLs already received.

Fixes #1088

## Review notes

- The change reuses the existing, tested `_format_image` download path rather than adding new logic; `data:` and `gs://` behavior is unchanged (both already returned base64 via `_format_image`).
- **Two existing unit tests asserted the old url-passthrough behavior (which encoded the bug); they now assert the downloaded-base64 result with the download mocked.** Please review those assertion changes specifically.
- No public API / signature change.

## Tests

`test_format_messages_tool_message_with_url_image_block` (rewritten as a #1088 regression test) and the url case in the parametrized `test_format_messages_anthropic` now mock `ImageBytesLoader.load_bytes` and assert a base64 source. Both fail before the change and pass after. Full `vertexai` unit suite: 298 passed, 2 skipped.

---

_Disclosure: Drafted with Claude Code. The root-cause analysis, fix, and regression tests were generated by an AI agent and reviewed and verified by me before submission._

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
@@ -423,7 +423,15 @@ def _clean_content_block(block: Any) -> Any:
     if block.get("type") not in ("tool_use", "image"):
         keys_to_remove.add("id")
 
-    return {k: v for k, v in block.items() if k not in keys_to_remove}
+    cleaned = {k: v for k, v in block.items() if k not in keys_to_remove}
+
+    # tool_result blocks nest their own content blocks, which can also carry
+    # streaming metadata (e.g. `index`). The top-level strip above does not
+    # reach them, so recurse into the nested content (#1256).
+    if block.get("type") == "tool_result" and isinstance(cleaned.get("content"), list):
+        cleaned["content"] = [_clean_content_block(b) for b in cleaned["content"]]
+
+    return cleaned
 
 
 def _clean_content(content: Any) -> Any:
--- a/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
@@ -358,14 +358,12 @@ def _format_image_content_block(block: dict, project: str | None = None) -> dict
         }
     if "url" in block:
         url = block["url"]
-        if url.startswith("data:"):
-            return {
-                "type": "image",
-                "source": _format_image(url, project),
-            }
+        # Claude on Vertex AI rejects url-type image sources (unlike Anthropic's
+        # direct API). Download and inline as base64 via _format_image, which
+        # already handles data:, http(s), and gs:// URLs. (#1088)
         return {
             "type": "image",
-            "source": {"type": "url", "url": url},
+            "source": _format_image(url, project),
         }
     if "file_id" in block:
         return {
--- a/libs/vertexai/tests/unit_tests/test_anthropic_utils.py
+++ b/libs/vertexai/tests/unit_tests/test_anthropic_utils.py
@@ -1226,6 +1226,49 @@ def test_format_messages_tool_message_with_streaming_metadata() -> None:
     }
 
 
+def test_format_messages_preformatted_tool_result_nested_streaming_metadata() -> None:
+    """Streaming metadata is stripped from nested content of pre-formatted tool_results.
+
+    Regression test for https://github.com/langchain-ai/langchain-google/issues/1256
+
+    When a ToolMessage already holds `tool_result` blocks (e.g. rehydrated from a
+    prior turn in a multi-agent graph), `_clean_content_block` stripped streaming
+    metadata from the top level of the `tool_result` block but left its nested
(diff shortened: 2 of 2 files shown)


chuenchen309:
CI note: the single red check is the live `llm-integration-tests` suite. This PR only changes the `type: "image"` content-block path (`_format_image_content_block`). The image tests in that suite — `test_pdf_gcs_uri`, `test_pdf_byts`, `test_https_image` — all go through the unchanged `type: "image_url"` → `_format_image` path, so the failure looks unrelated to this diff. The new base64-download behavior is covered by unit tests in `tests/unit_tests/test_anthropic_utils.py` (green). Happy to add an integration test exercising the `type: "image"` URL→base64 path if that'd be useful.

chuenchen309:
Following up on my CI note above — it was incomplete, and I'd rather correct it than let it stand.

Two problems with it:

1. I said this PR "only changes the `type: "image"` content-block path". It doesn't. It also makes `_clean_content_block` recurse into `tool_result` nested content (#1256), which touches every `tool_result` block, not just image ones. That's the part of the diff most likely to show up in a live suite, and my note didn't mention it at all.

2. I wrote that the failure "looks unrelated" without being able to read the build logs — they're in a private Cloud Build project. That was inference presented as triage, which isn't fair to you.

What I can actually verify: the three tests I named (`test_pdf_gcs_uri`, `test_pdf_byts`, `test_https_image` in `test_anthropic_files.py`) all send `{"type": "image_url", "image_url": {"url": ...}}`, which is a different branch from the `type: "image"` block this PR changes — so those three specifically shouldn't be affected by the image change. That's a much narrower claim than the one I made.

For what it's worth, the `vertexai` integration suite passes on my other PRs against this same file (#1884, #1887), so this doesn't look like a blanket fork-PR gate either.

If you can share which test failed, I'll fix it. I'm also happy to split the `tool_result` recursion out into its own PR — bundling it with the image fix here was a mistake on my part, and separating them would make this easier to judge.

PR #1889, opened 2026-07-11
fix(genai): don't mutate caller's tool dict in convert_to_genai_function_declarations

chuenchen309:
## Why

`_format_to_genai_function_declaration` binds `function` to the caller's tool dict (or its nested `"function"`) and then normalizes `parameters` **in place**:

```python
function = tool["function"]          # (or `cast("dict", tool)`)
function["parameters"] = function.get("parameters") or {}
if not function["parameters"].get("properties"):
    function["parameters"] = {}
```

So `convert_to_genai_function_declarations([tool])` mutates the caller's object:

- an OpenAI-format tool whose `parameters` lack a `properties` key has its `parameters` **wiped to `{}`** (losing `required`, `additionalProperties`, …);
- a parameterless `{"name", "description"}` dict gets a `parameters` key **added** to it.

Since tool dicts are frequently shared/reused across providers and calls, this silently corrupts the caller's data.

## What

Shallow-copy `function` before normalizing `parameters`. The lines only reassign the top-level `parameters` key, so a shallow copy fully prevents caller mutation while leaving the produced `FunctionDeclaration` byte-identical.

## Tests

Added `test_convert_to_genai_function_declarations_does_not_mutate_input` (OpenAI-format-without-properties and parameterless cases). It fails on `main` (the input dict is mutated) and passes with the fix. Full `test_function_utils.py`: **33 passed**; `ruff` and `mypy` clean.

---

🤖 *AI disclosure:* found by an automated code-review pass and fixed/tested with Claude Code (an AI coding agent). The regression is pinned by the new test.

### DIFF
--- a/libs/genai/langchain_google_genai/_function_utils.py
+++ b/libs/genai/langchain_google_genai/_function_utils.py
@@ -415,6 +415,9 @@ def _format_to_genai_function_declaration(
             function = convert_to_openai_tool(cast("dict", tool))["function"]
         else:
             function = cast("dict", tool)
+        # Copy so we don't mutate the caller's tool dict when normalizing
+        # `parameters` below (several branches above alias the input dict).
+        function = dict(function)
         function["parameters"] = function.get("parameters") or {}
         # Empty 'properties' field not supported.
         if not function["parameters"].get("properties"):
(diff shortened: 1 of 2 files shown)


chuenchen309:
CI note: the single red check is `langchain-google-genai-us (llm-integration-tests)`, the live-network integration suite. This change is confined to `_function_utils.py` — it shallow-copies the tool dict before normalizing `parameters`, producing a byte-identical `FunctionDeclaration` (covered by `test_function_utils.py`, all 33 green, plus the 12 non-integration checks here). So the integration red looks unrelated to this diff. Happy to dig in if a specific integration test is implicated.

PR #1895, opened 2026-07-17
feat(genai): add `reasoning_effort` as a standard chat model parameter

imnishitha:
### Summary

Adds support for the standard `reasoning_effort` parameter in `ChatGoogleGenerativeAI`. Like `temperature`, it can be passed at model construction or per call and is translated into Gemini's native `thinking_level`.

This extends the cross-provider `reasoning_effort` interface supported by other providers (langchain-ai/langchain#38887), allowing callers to configure reasoning consistently across Gemini as well. Supported values (`"minimal"`, `"low"`, `"medium"`, `"high"`) and the provider default are exposed via `ModelProfile` for Gemini 3.x models.

```python

model = ChatGoogleGenerativeAI(model="gemini-3-pro-preview")
model.invoke(
    "Why do parrots have colorful feathers?",
    reasoning_effort="high",
)
```

`thinking_level` is an alias for `reasoning_effort`; `reasoning_effort` remains the canonical field. The supported values match Gemini's native `thinking_level` exactly: `"minimal"`, `"low"`, `"medium"`, and `"high"`.

`ChatVertexAI` is not included because the underlying `google-cloud-aiplatform` SDK has no equivalent to `thinking_level`.

### Release note

`ChatGoogleGenerativeAI` now accepts `reasoning_effort`/ `thinking_level` on Gemini 3.x models. Supported values (`"minimal"`, `"low"`, `"medium"`, `"high"`) and the default are exposed via `ModelProfile`.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -2422,6 +2422,26 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
         `thinking_level` takes precedence over `thinking_budget` for Gemini 3+ models.
     """
 
+    reasoning_effort: Literal["none", "low", "medium", "high"] | None = Field(
+        default=None,
+    )
+    """Cross-provider reasoning effort setting for Gemini 3+ models.
+
+    Provides a standard `reasoning_effort` parameter consistent with other chat
+    model integrations (e.g. `ChatOpenAI`, `ChatAnthropic`).
+
+    Supported values:
+        * `'none'`: Disables thinking.
+        * `'low'`: Minimizes latency and cost.
+        * `'medium'`: Balances latency/cost with reasoning depth.
+        * `'high'`: Maximizes reasoning depth.
+
+    `reasoning_effort` is translated to Gemini's native thinking settings. Native
+    fields (`thinking_level`, `thinking_budget`, `thinking_config`) take precedence
+    if explicitly set. `'xhigh'`/`'max'` are intentionally unsupported since Gemini
+    has no equivalent.
+"""
+
     cached_content: str | None = None
     """The name of the cached content used as context to serve the prediction.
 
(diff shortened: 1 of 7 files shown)


libs/genai/langchain_google_genai/chat_models.py
```diff
@@ -2434,6 +2441,15 @@ class Joke(BaseModel):
 
     def __init__(self, **kwargs: Any) -> None:
         """Needed for arg validation."""
+        # `reasoning_effort` is a cross-provider-compatible alias for
+        # `thinking_level` (see that field's docstring). Resolved here, ahead of
+        # `model_fields`-based validation below and the `build_extra` validator
+        # (which only recognizes `Field(alias=...)`, not this manual alias), so
```
ccurme: why not use a pydantic alias?

libs/genai/tests/integration_tests/test_chat_models.py
```diff
@@ -353,6 +353,53 @@ def test_chat_google_genai_invoke_thinking(
...
+        **backend_config,
+    )
+    assert llm.thinking_level == reasoning_effort
+
+    result = llm.invoke(
+        "How many O's are in Google? Please tell me how you double checked the result",
+    )
+
+    assert isinstance(result, AIMessage)
+    if isinstance(result.content, list):
```
ccurme: let's make assertions on `result.content_blocks`

ccurme (commented):
LGTM (will need to remove your temporary install of langchain-core off git)

Thanks!

libs/genai/langchain_google_genai/chat_models.py
imnishitha: because `build_extra` checks fields before Pydantic alias handling, and only understands plain alias.

ccurme: do our existing aliases work? e.g., https://github.com/langchain-ai/langchain-google/blob/43432b8bc9cd1fb8b9daa06fc9301a09baf236bd/libs/genai/langchain_google_genai/chat_models.py#L2346-L2347

libs/genai/pyproject.toml
```diff
@@ -61,6 +61,13 @@ test_integration = [
 [tool.uv]
 constraint-dependencies = ["urllib3>=2.6.3"]
 
+[tool.uv.sources]
+# Temporary: pin to the langchain PR branch that adds reasoning_effort_levels/
+# reasoning_effort_default to ModelProfile (langchain-ai/langchain#38887), since
+# that hasn't released yet. Switch to a real langchain-core version pin once it
+# does.
+langchain-core = { git = "https://github.com/langchain-ai/langchain", branch = "nm/reasoning-effort", subdirectory = "libs/core" }
```
ccurme, line 69: LGTM once we release core and update here!

ccurme (approved):

PR #1913, opened 2026-07-30
fix(vertexai): use Vertex profiles

### LINKED ISSUE #1773: Models like `claude-haiku-4-5@20251001` are initialized with max output tokens 4096
### Package (Required)

- [ ] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python

location = "europe-west1"
model_name = "claude-haiku-4-5@20251001"
prompt = "Say Hi."

for model_name in ["claude-haiku-4-5@20251001", "claude-haiku-4-5-20251001"]:
    print(f"=== Test {model_name} ===")
    llm = ChatAnthropicVertex(
        model_name=model_name,
        location=location,
        max_retries=0,
    )

    print(f"max_output_tokens: {llm.max_output_tokens}")

    try:
        response = llm.invoke(prompt)
        print("LLM Call successful")
    except Exception as _:
        print("Error during LLM call")
    
    print()
```

```shell
$ uv run minimal_example_anthropic_vertexai_bug.py 
=== Test claude-haiku-4-5@20251001 ===
max_output_tokens: 4096
LLM Call successful

=== Test claude-haiku-4-5-20251001 ===
max_output_tokens: 64000
Error during LLM call
```

### Description

1. Initializing a model like `claude-haiku-4-5@20251001` (containing the `@` symbol) receives the default max token output settings (and possibly others). As this is the correct name of the model, the LLM executes correctly.

2. Initializing a model like `claude-haiku-4-5-20251001` (containing the `-` symbol) correctly gets the model-aware setting for max output tokens, but fails to execute, since the name is not correct.

---

I believe there is a model name mixup. VertexAI usually uses the `@` in the model name, Anthropic defaults to the `-` in the model names.

Compare https://github.com/langchain-ai/langchain-google/blob/main/libs/vertexai/langchain_google_vertexai/data/_profiles.py with https://github.com/langchain-ai/langchain-google/blob/main/libs/vertexai/langchain_google_vertexai/data/anthropic/_profiles.py

I suppose the bug is in this file and commit:
https://github.com/langchain-ai/langchain-google/commit/d0f74ae530f0bb278a4e83af85fba3ee6f8d7188#diff-36d4e123e72dba677fe8b97da0e287e47a3d676445f91ecce62828ff331cd2c6R244

gulbaki:
## Description

Use Google Vertex profiles for `ChatAnthropicVertex` so Vertex model names such as `claude-haiku-4-5@20251001` receive the correct model-aware token limit.

## Relevant issues

Fixes #1773  
Related to #1784 as a minimal fix based on the current `main` branch.

## Type
🐛 Bug Fix  
✅ Test

## Changes
- Use Vertex profiles for Anthropic model lookup.
- Update tests with valid Vertex model names.

## Testing
- 111 unit tests passed.
- Ruff, mypy, formatting, and import checks passed.


gulbaki:
Hi @lkuligin @mdrxy, could you please take a look when you have a chance?

This is a small, focused fix for #1773 that uses the existing Google Vertex profiles for `ChatAnthropicVertex`, along with updated unit tests.

mdrxy:
Would you be okay with the two latest follow-up commits on this branch?

They keep the bare Claude-name resolution, but make it safer by falling back when the shared Vertex registry has an ambiguous match, ignoring non-Claude profiles, and documenting the `@`-versioned Vertex-ID behavior. They also add focused regression coverage for those cases.

gulbaki:
> Would you be okay with the two latest follow-up commits on this branch?
> 
> They keep the bare Claude-name resolution, but make it safer by falling back when the shared Vertex registry has an ambiguous match, ignoring non-Claude profiles, and documenting the `@`-versioned Vertex-ID behavior. They also add focused regression coverage for those cases.

Yes, I’m okay with these changes.

PR #1933, opened 2026-08-08
fix(vertexai): convert v1 content blocks and stamp model_provider for Claude on Vertex

### LINKED ISSUE #1754: Tool use content blocks are "non-standard" in ChatAnthropicVertex responses
### Package (Required)

- [ ] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
import json

@tool
def echo_tool(input: str) -> str:
    """A simple tool that echoes the input."""
    return input


llm = (
    ChatAnthropicVertex(model="claude-sonnet-4-6", temperature=1, include_thoughts=True)
    .bind_tools([echo_tool])
    .bind(thinking={"type": "enabled", "budget_tokens": 1024})
)
response = llm.invoke("Use the echo tool to say hello world")

print(json.dumps(response.content_blocks, indent=2))
```

```shell

```

### Description

I am trying to get the content blocks of a response with thinking and tool use. The returned content blocks have the following issues:
* Thinking is formatted as a "non-standard" block, instead of a `ReasoningContentBlock`
* An extra non-standard content block representing the tool call is generated alongside the proper tool call block

### Actual output from snippet
```json
[
  {
    "type": "non_standard",
    "value": {
      "signature": "<thought signature>",
      "thinking": "The user wants me to use the echo tool to say \"hello world\".",
      "type": "thinking"
    }
  },
  {
    "type": "non_standard",
    "value": {
      "id": "toolu_vrtx_...",
      "caller": null,
      "input": {
        "input": "hello world"
      },
      "name": "echo_tool",
      "type": "tool_use"
    }
  },
  {
    "type": "tool_call",
    "id": "toolu_vrtx_...",
    "name": "echo_tool",
    "args": {
      "input": "hello world"
    }
  }
]
```

### Expected output
```json
[
  {
    "type": "reasoning",
    "reasoning": "The user wants me to use the echo tool to say \"hello world\".",
    "extras": {
      "signature": "<thought signature>"
    }
  },
  {
    "type": "tool_call",
    "name": "echo_tool",
    "args": {
      "input": "hello world"
    },
    "id": "toolu_vrtx_..."
    "extras": {
      "caller": null
    }
  }
]
```

### Attempt to fix
The problem seems to arise from [`langchain_core.messages.ai.AIMessage.content_blocks`](https://github.com/langchain-ai/langchain/blob/d7031101da78e3f6b6c5956b48a5170c1a33292b/libs/core/langchain_core/messages/ai.py#L253) not being able to find the `model_provider` metadata, and thus not parsing content using the Anthropic content block translator.   
Adding

```python
response.response_metadata = {"model_provider": "anthropic"}
```

to the above code before the print statement leads to the expected output.

### LINKED ISSUE #1930: ChatAnthropicVertex: _format_messages_anthropic does not convert v1 content blocks back, so replaying a tool call fails with 400 invalid_request
### Package (Required)

- [ ] langchain-google-genai

### Checked other resources

- [x] I searched the LangChain documentation and API reference

### Description

`_format_messages_anthropic` cannot read most of the standard v1 content blocks. It handles `reasoning` (added for #1654) but forwards every other v1-only block type unchanged into the request payload, where Anthropic rejects it with `400 invalid_request`.

The affected block types are `tool_call`, `tool_call_chunk`, `non_standard`, `server_tool_call` and `server_tool_result`. `langchain-anthropic` translates all of them back in [`_compat.py::_convert_from_v1_to_anthropic`](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/_compat.py); the Vertex integration adopted only the `reasoning` case.

The relevant lines are in `_anthropic_utils.py::_format_message_anthropic`: `reasoning` is converted at lines 223-235, and anything unrecognised reaches the catch-all `content.append(block)` at line 257.

This is easy to miss with a plain `.invoke()` because content stays provider-native there. It is unavoidable as soon as content is standardised, which happens with `output_version="v1"`, with `LC_OUTPUT_VERSION=v1`, and — silently — for any caller that attaches a `_V2StreamingCallbackHandler`, since `_should_use_protocol_streaming` rewrites every chunk into v1 blocks regardless of `output_version`. Live token streaming in an agent is exactly that case.

### Example Code (Python)

```python
from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
from langchain_google_vertexai._anthropic_utils import _format_messages_anthropic

# What langchain-core produces once content is standardised.
ai_message = AIMessage(
    content=[
        {
            "type": "tool_call",
            "id": "toolu_vrtx_01",
            "name": "reference_lookup",
            "args": {"query": "test"},
        }
    ],
    tool_calls=[
        {
            "type": "tool_call",
            "id": "toolu_vrtx_01",
            "name": "reference_lookup",
            "args": {"query": "test"},
        }
    ],
)

_, formatted = _format_messages_anthropic(
    [
        HumanMessage("test"),
        ai_message,
        ToolMessage(content="result", tool_call_id="toolu_vrtx_01", name="reference_lookup"),
    ],
    "my-project",
)
print(formatted[1]["content"])
```

### Actual output

The v1 `tool_call` block is passed through verbatim, next to the correct `tool_use` block the formatter appends from `message.tool_calls`:

```python
[
  {'type': 'tool_call', 'name': 'reference_lookup', 'args': {'query': 'test'}},
  {'type': 'tool_use', 'name': 'reference_lookup', 'input': {'query': 'test'}, 'id': 'toolu_vrtx_01'},
]
```

### Expected output

```python
[
  {'type': 'tool_use', 'name': 'reference_lookup', 'input': {'query': 'test'}, 'id': 'toolu_vrtx_01'},
]
```

```shell
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': "messages.1.content.0.type: Input tag 'tool_call' found using 'type' does not match any of the expected tags: 'text', 'image', 'document', 'search_result', 'thinking', 'redacted_thinking', 'tool_use', 'tool_result', ..."}}
```

With `non_standard` blocks — which is what you get today whenever `model_provider` is absent, see #1754 — the same call fails with `Input tag 'non_standard' ...` instead.

### Suggested fix

Mirror `langchain_anthropic._compat._convert_from_v1_to_anthropic` in `_format_message_anthropic`, ahead of the `content.append(block)` catch-all. The `tool_use` de-duplication already present at lines 247-255 then removes the converted block again and rebuilds it from `message.tool_calls`, so the fix is additive and does not change existing behaviour for provider-native content.

Note that this and #1754 are two halves of one failure for streaming callers: without `model_provider` the tool call arguments are already lost before the formatter is reached, and fixing only `model_provider` still leaves the request rejected here. Both are needed for a streaming agent to complete a tool call against Claude on Vertex.

### System Info

`langchain-google-vertexai` 3.2.4, `langchain-core` 1.5.1, `langchain` 1.3.11, `anthropic` 0.120.0, Python 3.13, `claude-sonnet-5` on Vertex in `eu`.

Nikshaan:
## Description

Fixes two related bugs in ChatAnthropicVertex when handling
LangChain v1 content blocks:

1. `model_provider` was never set on response/stream metadata, so
`AIMessage.content_blocks` had no Anthropic translator and native
content degraded to `non_standard` blocks. On the streaming path this
was worse: `non_standard` blocks were replaced rather than concatenated
on chunk merge, so streamed tool-call arguments collapsed to the last
JSON fragment instead of the full payload.
2. `_format_message_anthropic` only converted `reasoning` blocks back to
Anthropic-native shape. Other v1-only types (`tool_call`,
`tool_call_chunk`, `non_standard`, `server_tool_call`,
`server_tool_result`) fell through to the catch-all and were sent to
the API unchanged, which Anthropic rejected with a 400.

Together these made tool calling impossible for any agent using
streaming Claude on Vertex, since replaying a tool result after a streamed
call would trigger both issues simultaneously.

Fix:
- Set `response_metadata["model_provider"] = "anthropic"` on
`_format_output` (all branches) and at the end of
`_make_message_chunk_from_anthropic_event`, as
`langchain_anthropic` does.
- Add `_convert_v1_block_to_anthropic` to convert the affected v1 types
before the type dispatch in `_format_message_anthropic`, so a converted
`tool_call` would be deduped against `message.tool_calls` by the
existing `tool_use` logic, with no need for new dedup code.
- Fix a related bug this exposed: `_clean_content_block` was stripping
  `id` from `tool_call`/`tool_call_chunk` blocks (its allowlist only
  included native `tool_use`/`image`), which defeated the new dedup
  logic — a converted block with `id=""` was never considered a
  duplicate. This manifested as a test failure until the fix landed.

All new dict-building code is non-mutating; content blocks may be shared
with agent state/checkpoints so in-place edits would have been unsafe.

## Relevant issues

Fixes #1754
Fixes #1930

## Type

🐛 Bug Fix
✅ Test

## Changes(optional)

- `_anthropic_utils.py`: new `_convert_v1_block_to_anthropic`; call site
in `_format_message_anthropic`; `model_provider` stamp in
`_make_message_chunk_from_anthropic_event`; `_clean_content_block`
allowlist extended to keep `id` for `tool_call`/`tool_call_chunk`.
- `model_garden.py`: `model_provider` stamp in `_format_output` (all
three branches).

## Testing(optional)

Added unit tests covering: v1 `tool_call` round-trip with no duplicate
block, validation against Anthropic SDK's `ContentBlockParam`,
`tool_call_chunk` partial-JSON fallback, `non_standard` unwrapping with
nested streaming keys stripped, non-mutation of input messages, and
`model_provider` on all four streaming event types. Verified each new/
modified assertion fails on the pre-fix code and passes post-fix.

`make lint` (ruff, ruff format, mypy) and full unit test suite for
`libs/vertexai` pass.

## Note(optional)

Two `server_tool_call`/`server_tool_result` branches were ported
alongside the reported types since they hit the same catch-all gap, but
have no dedicated test - no local repro path exists for
`code_interpreter`/`remote_mcp` blocks without live server-tool
responses. Happy to add coverage if useful.

### DIFF
--- a/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
@@ -1,4 +1,5 @@
 import base64
+import json
 import re
 import urllib.parse
 import warnings
@@ -142,6 +143,90 @@ def _format_text_content(text: str) -> dict[str, str | dict[str, Any]]:
     return content
 
 
+def _convert_v1_block_to_anthropic(block: dict) -> dict:
+    """Convert a LangChain v1 content block to Anthropic wire format.
+
+    Builds new dicts and never mutates `block` in place. Unrecognized types are
+    returned unchanged. `reasoning` is intentionally not handled here — the
+    dedicated branch in `_format_message_anthropic` supports `cache_control`.
+
+    Args:
+        block: A content block dict, possibly in LangChain v1 shape.
+
+    Returns:
+        An Anthropic-native content block dict, or the original block when no
+        conversion applies.
+    """
+    block_type = block.get("type")
+
+    if block_type == "tool_call":
+        return {
+            "type": "tool_use",
+            "name": block.get("name", ""),
+            "input": block.get("args", {}),
+            "id": block.get("id", ""),
+        }
+
+    if block_type == "tool_call_chunk":
+        args = block.get("args")
+        if isinstance(args, str):
+            try:
+                input_ = json.loads(args or "{}")
+            except json.JSONDecodeError:
+                input_ = {}
+        else:
+            input_ = args or {}
+        return {
+            "type": "tool_use",
+            "name": block.get("name", ""),
+            "input": input_,
+            "id": block.get("id", ""),
+        }
+
+    if block_type == "non_standard" and "value" in block:
+        # `_clean_content` only strips streaming keys at the top level, so
+        # nested `index` / `partial_json` / `caller` inside `value` survive
+        # merge. Clean the unwrapped block before returning.
+        return _clean_content_block(block["value"])
+
+    if block_type == "server_tool_call":
+        new_block: dict[str, Any] = {}
+        if "id" in block:
+            new_block["id"] = block["id"]
+        new_block["input"] = block.get("args", {})
+        if partial_json := block.get("extras", {}).get("partial_json"):
+            new_block["input"] = {}
+            new_block["partial_json"] = partial_json
+        if block.get("name") == "code_interpreter":
+            new_block["name"] = "code_execution"
+        elif block.get("name") == "remote_mcp":
+            if "tool_name" in block.get("extras", {}):
+                new_block["name"] = block["extras"]["tool_name"]
+            if "server_name" in block.get("extras", {}):
+                new_block["server_name"] = block["extras"]["server_name"]
+        else:
+            new_block["name"] = block.get("name", "")
+        if block.get("name") == "remote_mcp":
+            new_block["type"] = "mcp_tool_use"
+        else:
+            new_block["type"] = "server_tool_use"
+        return new_block
+
+    if block_type == "server_tool_result":
+        new_block = {}
+        if "output" in block:
+            new_block["content"] = block["output"]
+        server_tool_result_type = block.get("extras", {}).get("block_type", "")
+        if server_tool_result_type == "mcp_tool_result":
+            new_block["is_error"] = block.get("status") == "error"
+        if "tool_call_id" in block:
+            new_block["tool_use_id"] = block["tool_call_id"]
+        new_block["type"] = server_tool_result_type
+        return new_block
+
+    return block
+
+
 def _format_message_anthropic(
     message: HumanMessage | AIMessage | SystemMessage, project: str | None
 ):
@@ -179,6 +264,8 @@ def _format_message_anthropic(
                 if "type" not in block:
                     msg = "Dict content block must have a type key"
                     raise ValueError(msg)
+
+                block = _convert_v1_block_to_anthropic(block)
 
                 new_block = {}
 
@@ -418,9 +505,9 @@ def _clean_content_block(block: Any) -> Any:
     # 'partial_json' - added during streaming for incremental JSON parsing
     keys_to_remove = {"index", "partial_json", "caller"}
 
-    # The id field is required for tool_use blocks and some image blocks,
-    # but forbidden in text blocks (specifically inside tool_results).
-    if block.get("type") not in ("tool_use", "image"):
+    # Keep id for tool_use/image and v1 tool_call(s) — format dedup needs
+    # that id after conversion. Strip it from text (esp. inside tool_results).
+    if block.get("type") not in ("tool_use", "image", "tool_call", "tool_call_chunk"):
         keys_to_remove.add("id")
 
     return {k: v for k, v in block.items() if k not in keys_to_remove}
@@ -630,6 +717,8 @@ def _make_message_chunk_from_anthropic_event(
         )
     else:
         pass
+    if message_chunk:
+        message_chunk.response_metadata["model_provider"] = "anthropic"
     return message_chunk
 
 
--- a/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
+++ b/libs/vertexai/langchain_google_vertexai/_anthropic_utils.py
@@ -160,12 +160,15 @@ def _convert_v1_block_to_anthropic(block: dict) -> dict:
     block_type = block.get("type")
 
     if block_type == "tool_call":
-        return {
+        tool_use_block = {
             "type": "tool_use",
             "name": block.get("name", ""),
             "input": block.get("args", {}),
             "id": block.get("id", ""),
         }
+        if "caller" in block.get("extras", {}):
+            tool_use_block["caller"] = block["extras"]["caller"]
+        return tool_use_block
 
     if block_type == "tool_call_chunk":
         args = block.get("args")
@@ -338,7 +341,7 @@ def _format_message_anthropic(
                         is_unique = block["id"] not in [
                             tc["id"] for tc in message.tool_calls
                         ]
-                        if not is_unique:
+                        if not is_unique and not block.get("caller"):
                             continue
 
                 content.append(block)
@@ -347,7 +350,15 @@ def _format_message_anthropic(
         raise ValueError(msg)
 
     if isinstance(message, AIMessage) and message.tool_calls:
-        for tc in message.tool_calls:
+        tool_use_ids = [
+            b["id"]
+            for b in content
+            if isinstance(b, dict) and b.get("type") == "tool_use"
+        ]
+        missing_tool_calls = [
+            tc for tc in message.tool_calls if tc["id"] not in tool_use_ids
+        ]
+        for tc in missing_tool_calls:
             tu = cast("dict[str, Any]", _lc_tool_call_to_anthropic_tool_use_block(tc))
             content.append(tu)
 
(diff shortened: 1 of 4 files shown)


Nikshaan:
Per the contributing guide, could you formally assign me to #1754 and #1930? Want to make sure this doesn't get auto-closed for missing that step.

nicolas-ziegler-bayard:
Nice fix — this covers everything from the issue discussion cleanly (ordering, the `id`-stripping trap that broke dedup, `non_standard` unwrap with nested-key cleanup, `tool_call_chunk` partial-JSON fallback, and stamping `model_provider` on both the output and streaming paths for #1754). Verified against upstream `langchain-anthropic` and it matches `_compat.py`'s conversion shapes closely.

One gap I think is still open: the `caller` round-trip for programmatic tool use (Claude's `code_execution` calling a tool directly via `allowed_callers`) isn't preserved.

**1. `_convert_v1_block_to_anthropic`'s `tool_call` branch drops `extras.caller`**

```python
if block_type == "tool_call":
    return {
        "type": "tool_use",
        "name": block.get("name", ""),
        "input": block.get("args", {}),
        "id": block.get("id", ""),
    }
```

`langchain-anthropic`'s equivalent (`_compat.py:115-124`) explicitly carries it over:

```python
elif block["type"] == "tool_call":
    tool_use_block = {
        "type": "tool_use",
        "name": block.get("name", ""),
        "input": block.get("args", {}),
        "id": block.get("id", ""),
    }
    if "caller" in block.get("extras", {}):
        tool_use_block["caller"] = block["extras"]["caller"]
    new_content.append(tool_use_block)
```

`caller` lands in `extras` on the v1 side (see `langchain_core`'s `block_translators/anthropic.py:291-294`), not at the top level, so `_clean_content_block`'s unconditional `caller` strip doesn't touch it before this conversion runs — it really is being dropped here.

**2. The dedup check still lacks the `caller` exception**

```python
if isinstance(message, AIMessage) and message.tool_calls:
    is_unique = block["id"] not in [
        tc["id"] for tc in message.tool_calls
    ]
    if not is_unique:
        continue

content.append(block)
```

`langchain-anthropic`'s `_format_messages` only takes this "prefer the `tool_calls` version" path `and not block.get("caller")` (`chat_models.py:565-569`); when `caller` is set it falls through to rebuilding the block from `block` itself, which is what actually preserves `caller`. Without that exception, a `caller`-tagged `tool_use` whose id also appears in `message.tool_calls` gets dropped here and replaced later by a caller-less block built from `message.tool_calls`.

**Why this is reachable now:** before this PR, a `caller`-tagged native block just degraded to `non_standard` since `model_provider` was never set, so nobody looked at `caller`. Now that `model_provider="anthropic"` is stamped everywhere, `AIMessage.content_blocks` will run the real Anthropic v1 translator and produce `tool_call`/`extras.caller` for these blocks — so this is now a live path, not a theoretical one. `langchain-anthropic` has an integration test that covers exactly this (`test_programmatic_tool_use`, asserting `"caller" in tool_call_block["extras"]`), which is worth mirroring here for regression coverage since `code_execution` is already a supported server tool on this Vertex integration.

Suggested fix: mirror `_compat.py`'s `if "caller" in block.get("extras", {})` copy in the `tool_call` branch of `_convert_v1_block_to_anthropic`, and add the matching `and not block.get("caller")` guard to the `tool_use` dedup check in `_format_message_anthropic`.

Nikshaan:
Confirmed both points against the code. `caller` does land in `extras`, not top-level, and survives `_clean_content_block` untouched since that only strips top-level keys. It was being dropped at the conversion step, exactly as you found.

Fixed the conversion and added the dedup exception, but that surfaced a third issue: unlike `langchain_anthropic`'s `missing_tool_calls` pattern, our final loop appends every `message.tool_calls` entry unconditionally. So once the dedup lets a caller-tagged block through, that loop added a caller-less duplicate for the same id. Ported the same `tool_use_ids`/`missing_tool_calls` filtering.

Added a unit test adapted from `test_programmatic_tool_use`'s intent that one's a live integration test with a VCR cassette, so I wrote a unit-level equivalent asserting `caller` survives to the output and that exactly one `tool_use` block exists for the id. Verified it fails without the fix and passes with it.

nicolas-ziegler-bayard:
thanks alot @Nikshaan !
do you know when PRs like yours will be merged usally?

Nikshaan:
No idea on the usual timeline, but hopefully it moves along once a maintainer gets a look.

nicolas-ziegler-bayard:
any updates?

Nikshaan:
Hey, no updates from my side. The last PR merged into the repo was about a week ago, so review bandwidth seems a bit limited right now. If you're comfortable doing so, would you mind flagging this to a maintainer, whenever it's convenient?

nicolas-ziegler-bayard:
@artdent @nfcampos @ninjapenguin  
I hope you could have a look at this, as i assume you are the maintianers? Thanks :)

Nikshaan:
@mdrxy Following up. This has been up for review for about 3 weeks now and the assignment from my original comment hasn't been completed either yet. There's no rush if you have been busy, just wanted to make sure this hasn't been overlooked.

I would also be happy to add tests for the server_tool_call/server_tool_result branches if this helps things get moving.

Issue #1497, opened 2026-01-08
Test `thinkingConfig`/`thinkingLevel` compatibility


mdrxy:
### Package (Required)

- [x] langchain-google-genai
- [ ] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
"thinkingConfig": {
    "thinkingLevel": "low"
}
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

We need to test/ensure that `thinkingLevel` is accepted like `thinking_level`. Also what happens when using it wrapped in a `thinkingConfig`?

https://ai.google.dev/gemini-api/docs/thinking#rest_1

1Ninad:
Hi @mdrxy, is this resolved yet? I would like to work on it.

Issue #1366, opened 2025-11-19
thinking_level not accepted when initializing Gemini 3 via init_chat_model()


Gu-Haojia:
### Package (Required)

- [x] langchain-google-genai
- [ ] langchain-google-vertexai
- [ ] langchain-google-community
- [ ] Other / not sure / general

### Checked other resources

- [x] I added a descriptive title to this issue
- [x] I searched the LangChain documentation and API reference (linked above)
- [x] I used the GitHub search to find a similar issue and didn't find it
- [x] I am sure this is a bug and not a question or request for help

### Example Code (Python)

```python
from langchain.chat_models import init_chat_model

model = init_chat_model(
    "google_genai:gemini-3-pro-preview",
    thinking_level="low",  # <-- new field from Gemini docs
)

# do invoke
```

### Error Message and Stack Trace (if applicable)

```shell
Model return: Unknown field for ThinkingConfig: thinking_level
```

### Description

According to Google’s latest Gemini thinking docs (https://ai.google.dev/gemini-api/docs/thinking), thinking_level is a supported parameter alongside thinking_budget. When we attempt to pass thinking_level="low" through LangChain’s init_chat_model, the config validation rejects it with “Unknown field for ThinkingConfig: thinking_level”. This prevents us from using the new low/high thinking presets while still leveraging LangChain’s tooling/graph integrations.

Environment: langchain==1.0.7, langchain-core==1.0.5, langchain-google-genai==3.1.0.
Expectation: LangChain should either accept thinking_level and forward it to Gemini, or clearly document that only the legacy thinking_budget* fields are supported.

Yuki-Imajuku:
Same here.
This is because langchain-gogle-genai uses [google-ai-generativelanguage](https://github.com/googleapis/google-cloud-python/tree/main/packages/google-ai-generativelanguage), not [google-genai](https://github.com/googleapis/python-genai)

FedericoCampe8:
cc @ccurme possibly related to this? https://github.com/langchain-ai/langchain-google/pull/1365

Yuki-Imajuku:
I noticed someone address migration to `google-genai`. #1330

mbn18:
I tested it:

```
llm = ChatGoogleGenerativeAI(model='gemini-3-pro', media_resolution=3, thinking_level="low")
config = llm._prepare_params(stop=None)
```

Receiving the following error:
```
ValueError: Unknown field for ThinkingConfig: thinking_level
```

[...]

PR #1940, opened 2026-08-12
fix(genai): read model_kwargs when building the request config

### LINKED ISSUE #1938: genai: model_kwargs is never read, so unknown constructor kwargs are silently dropped
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

- [x] I searched the LangChain documentation and API reference

### Example Code (Python)

```python
import warnings

MSG = [HumanMessage("hi")]

for kwargs in [{"response_logprobs": True}, {"logprobs": 3}, {"audio_timestamp": True}]:
    name = next(iter(kwargs))
    model = ChatGoogleGenerativeAI(model="gemini-3.5-flash", api_key="k", **kwargs)
    from_ctor = getattr(model._prepare_request(MSG)["config"], name)
    from_invoke = getattr(model._prepare_request(MSG, **kwargs)["config"], name)
    print(name, "| model_kwargs:", model.model_kwargs, "| ctor:", from_ctor, "| invoke:", from_invoke)
```

```
response_logprobs | model_kwargs: {'response_logprobs': True} | ctor: None | invoke: True
logprobs          | model_kwargs: {'logprobs': 3}             | ctor: None | invoke: 3
audio_timestamp   | model_kwargs: {'audio_timestamp': True}   | ctor: None | invoke: True
```

No error. A warning, then the value is discarded:

```
WARNING! response_logprobs is not default parameter.
                response_logprobs was transferred to model_kwargs.
                Please confirm that response_logprobs is what you intended.
```

### Description

`ChatGoogleGenerativeAI` declares `model_kwargs` and populates it through `_build_model_kwargs`, but never reads it back when building a request. There are three references to it in `chat_models.py`, and none of them is a read:

- line 2362, the field declaration
- line 2546, `_build_model_kwargs(values, all_required_field_names)` in the validator
- the import at line 99

So any constructor keyword that is not a declared field warns and then vanishes, while the same keyword passed at invoke time works, because unknown invoke kwargs survive in `remaining_kwargs` and are spread into `GenerateContentConfig`. The two entry points disagree, and the one that silently does nothing is the one people reach for first.

The three fields above are just handy examples; the behavior is not specific to them. It applies to every `GenerateContentConfig` field without a corresponding LangChain field.

This is the root cause behind several reports that have been fixed one field at a time:

- #1497 / #1869, `thinking_config` "was swept into `model_kwargs` and never used"
- #1366, `thinking_level`
- #1682, `service_tier` (#1937)

Each fix promotes one more parameter to a real field, which works, but the next unlisted parameter fails the same way, and it fails quietly.

Two possible directions, and I don't know which you'd prefer:

1. Spread `model_kwargs` into the request config the way the OpenAI integration does, so the documented escape hatch actually reaches the API.
2. Keep dropping unknown keywords, but raise instead of warning, so a typo or an unsupported parameter is loud at construction time rather than a silent no-op in production.

Either is a small change. Option 1 also makes the `model_kwargs` field mean what its name suggests. Happy to send a PR for whichever you'd rather have.

### System Info

`langchain-google-genai` 4.3.3, `google-genai` 2.17.0, `langchain-core` 1.5.3, Python 3.11, macOS.

axiom-of-choice:
Closes #1938.

## Why

`ChatGoogleGenerativeAI` declares `model_kwargs` and fills it through `_build_model_kwargs`, but nothing ever read it back. There were three references in `chat_models.py` and none was a read. So a constructor keyword that is not a declared field warned and then vanished, while the same keyword passed at invoke time worked, because unknown invoke kwargs survive in `remaining_kwargs` and get spread into `GenerateContentConfig`.

The two entry points disagreed, and the one that silently did nothing is the one people reach for first. It is also the root cause behind several reports that were each fixed by promoting one more parameter to a real field (#1497 / #1869 `thinking_config`, #1366 `thinking_level`, #1682 `service_tier`). Every fix works, and then the next unlisted parameter fails the same quiet way.

## What changed

One line at the top of `_prepare_request`:

```python
kwargs = {**self.model_kwargs, **kwargs}
```

`model_kwargs` now takes the same path as invoke-time kwargs, so it flows through `_prepare_params`, the `_consumed_kwargs` filter and `remaining_kwargs` into `_build_request_config`. Per-call values win over the ones fixed at construction.

The issue offered two directions and asked which you preferred. This implements the first (spread `model_kwargs` into the config, matching the OpenAI integration). Happy to switch to the second (raise at construction instead of warning) if you would rather go that way.

Worth noting the two converge more than expected: because `GenerateContentConfig` forbids extra inputs, an unsupported key now raises `ValidationError` instead of being dropped. That is the same error the invoke path already raised for the same key, so this is the constructor catching up to existing behavior rather than new behavior being invented. Test `test_unknown_model_kwargs_raise_like_invoke_kwargs` pins both halves.

## Verification

Before, using the repro from the issue:

```
response_logprobs | model_kwargs: {'response_logprobs': True} | ctor: None | invoke: True
logprobs          | model_kwargs: {'logprobs': 3}             | ctor: None | invoke: 3
audio_timestamp   | model_kwargs: {'audio_timestamp': True}   | ctor: None | invoke: True
```

After:

```
response_logprobs | model_kwargs: {'response_logprobs': True} | ctor: True | invoke: True
logprobs          | model_kwargs: {'logprobs': 3}             | ctor: 3    | invoke: 3
audio_timestamp   | model_kwargs: {'audio_timestamp': True}   | ctor: True | invoke: True
```

Mutation check: reverting the source hunk and keeping the tests gives 5 failed, 1 passed. Restoring it gives 6 passed. The one that passes either way is the pre-existing `test_model_kwargs`, which only asserts on construction.

Full run: `387 passed` in `libs/genai` unit tests, `make lint` clean (ruff, format, import order, mypy on 35 source files).

## Risks

The behavior change worth reviewing is the one above: a model constructed today with a key that `GenerateContentConfig` does not accept currently ignores it at request time, and after this it raises. Nothing in the existing suite depended on that, which is why 387 tests pass untouched, but it is a real change for anyone parking arbitrary values in `model_kwargs`. If you want it strictly additive I can filter to keys `GenerateContentConfig` accepts and keep warning on the rest, though that reintroduces the silent drop this issue is about.

Scope is `libs/genai` only, matching the issue's package checkbox. `vertexai` has its own `model_kwargs` handling and is not touched here.

## Pre-merge

Nothing required. No new dependencies, no public signature changes.

## Post-merge

The `service_tier` promotion in #1937 stays useful as a typed field, and does not conflict with this.

---

Part of the work was done using an AI coding assistant, but tests and quality were manually reviewed.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -3026,6 +3026,12 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
         **kwargs: Any,
     ) -> dict[str, Any]:
         """Prepare the request configuration for the API call."""
+        # Constructor keywords that aren't declared fields are collected into
+        # `model_kwargs` by `_build_model_kwargs`. Merge them in here so they take the
+        # same path as invoke-time kwargs and reach `GenerateContentConfig`; per-call
+        # values win over the ones fixed at construction.
+        kwargs = {**self.model_kwargs, **kwargs}
+
         # Process tools and functions
         formatted_tools = self._format_tools(tools, functions)
 
--- a/libs/genai/tests/unit_tests/test_chat_models.py
+++ b/libs/genai/tests/unit_tests/test_chat_models.py
@@ -1564,6 +1564,65 @@ def test_model_kwargs(mock_client: Mock) -> None:
 
 
 @pytest.mark.parametrize(
+    "param,value",
+    [
+        ("response_logprobs", True),
+        ("logprobs", 3),
+        ("audio_timestamp", True),
+    ],
+)
+def test_model_kwargs_reach_request_config(param: str, value: Any) -> None:
+    """`model_kwargs` should configure the request, like the same kwarg at invoke."""
+    msg = HumanMessage(content="test")
+    llm = ChatGoogleGenerativeAI(
+        model=MODEL_NAME,
+        google_api_key=SecretStr(FAKE_API_KEY),
+        model_kwargs={param: value},
+    )
+
+    from_constructor = getattr(llm._prepare_request([msg])["config"], param)
+    invoked = llm._prepare_request([msg], **{param: value})
+    from_invoke = getattr(invoked["config"], param)
+
+    assert from_constructor == value
+    assert from_constructor == from_invoke
+
+
+def test_invoke_kwargs_override_model_kwargs() -> None:
+    """A per-call value wins over the one fixed at construction."""
+    msg = HumanMessage(content="test")
+    llm = ChatGoogleGenerativeAI(
+        model=MODEL_NAME,
+        google_api_key=SecretStr(FAKE_API_KEY),
+        model_kwargs={"logprobs": 3},
+    )
+
+    assert llm._prepare_request([msg], logprobs=7)["config"].logprobs == 7
+    # The override is per-call, so it must not leak into the next request.
+    assert llm._prepare_request([msg])["config"].logprobs == 3
+    assert llm.model_kwargs == {"logprobs": 3}
+
+
+def test_unknown_model_kwargs_raise_like_invoke_kwargs() -> None:
+    """An unsupported key fails the same way from either entry point."""
+    msg = HumanMessage(content="test")
+    llm = ChatGoogleGenerativeAI(
+        model=MODEL_NAME,
+        google_api_key=SecretStr(FAKE_API_KEY),
+        model_kwargs={"not_a_real_param": 1},
+    )
+    with pytest.raises(ValidationError, match="not_a_real_param"):
+        llm._prepare_request([msg])
+
+    plain = ChatGoogleGenerativeAI(
+        model=MODEL_NAME,
+        google_api_key=SecretStr(FAKE_API_KEY),
+    )
+    with pytest.raises(ValidationError, match="not_a_real_param"):
+        plain._prepare_request([msg], not_a_real_param=1)
+
+
+@pytest.mark.parametrize(
     "is_async,method_name,client_method",
     [
         (False, "_generate", "models.generate_content"),  # Sync


Daniil-Pankieiev:
The underlying problem here is real (#1938): `model_kwargs` is collected and then never read, so anything put there is silently dropped.

One concern with the merge as written, though. `kwargs = {**self.model_kwargs, **kwargs}` forwards *everything* in `model_kwargs` into `GenerateContentConfig`, and that model is declared with `extra='forbid'`. `_build_model_kwargs` collects **any** unrecognised constructor keyword into `model_kwargs`, including a plain typo, which today only produces a warning. After this change the same typo becomes a hard `ValidationError` on the first request.

Reproduction:

```python
import warnings
from pydantic import SecretStr
from google.genai.types import GenerateContentConfig

llm = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash",
    google_api_key=SecretStr("not-a-real-key"),
    temperatur=0.5,          # typo for `temperature`
)
print("model_kwargs:", llm.model_kwargs)

# today: the request still builds
cfg = llm._prepare_request([HumanMessage(content="hi")])["config"]
print("current: request built OK ->", type(cfg).__name__)

# what this PR's merge line produces
GenerateContentConfig(**{**llm.model_kwargs})
```

Output:

```
Unexpected argument 'temperatur' provided to ChatGoogleGenerativeAI. Did you mean: 'temperature'?
model_kwargs: {'temperatur': 0.5}
current: request built OK -> GenerateContentConfig
ValidationError: 1 validation error for GenerateContentConfig
temperatur
  Extra inputs are not permitted [type=extra_forbidden, input_value=0.5, input_type=float]
```

So the failure mode moves from a warning at construction time, which even suggests the right spelling, to an exception at request time, by which point that hint is gone and the traceback points at `GenerateContentConfig` rather than at the caller's constructor.

A couple of options that keep the fix and avoid the regression:

- Forward only the keys `GenerateContentConfig` actually accepts, for example intersecting with `GenerateContentConfig.model_fields`, and leave anything else on the existing warning path.
- Or validate at construction time instead, so an unknown key fails immediately with the "Did you mean" message rather than on the first call.

Either way it would be worth a test covering an unknown key in `model_kwargs`, since that is the path that changes behaviour here.

Tested against `langchain-google-genai` 4.3.5 with `google-genai` 2.19.0. I checked the current `master` too: `model_kwargs` is still only referenced at the import, the field declaration and the validator, and `_build_request_config` still splats `**kwargs` into `GenerateContentConfig`, so the same reasoning applies there.

PR #1953, opened 2026-08-19
fix(genai): populate token_usage in llm_output

### LINKED ISSUE #957: [BUG] ChatGoogleGenerativeAI does not populate usage_metadata in LLMResult.llm_output
When using `ChatGoogleGenerativeAI` from the `langchain-google-genai` library, the `LLMResult.llm_output` field is not being populated with the `usage_metadata` from the Google API response. It returns as `None`, an empty dictionary, or a dictionary containing only `prompt_feedback`.

This makes it impossible to track token usage (prompt tokens, completion tokens, total tokens) via standard LangChain mechanisms like `BaseCallbackHandler`, as the callback receives an `LLMResult` with no usage data.

This issue occurs in both standard chain invocations (`chain.invoke()`) and direct calls to the chat model instance (`llm.generate()`).

### Environment:
- Python Version: 3.12
- Library Versions:
  - `langchain`: 0.3.25
  - `langchain-core`: 0.3.60
  - `langchain-google-genai`: 2.0.10
  - `google-generativeai`: 0.8.4

### Steps to Reproduce
The following minimal, self-contained script reproduces the issue reliably. It uses the `ChatGoogleGenerativeAI` wrapper and calls the `.generate()` method directly.
```python
import os
import logging
from typing import List

import google.generativeai as genai
from langchain_core.messages import HumanMessage, BaseMessage
from langchain_core.outputs import LLMResult

# --- Basic Setup ---
load_dotenv()
logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')

# --- LLM Initialization ---
try:
    # Use a standard, publicly available model for the test
    llm = ChatGoogleGenerativeAI(
        model="gemini-1.5-flash-latest",
        google_api_key=os.getenv("GOOGLE_API_KEY") # Ensure this environment variable is set
    )
    logging.info("ChatGoogleGenerativeAI model initialized successfully.")
except Exception as e:
    logging.error(f"Failed to initialize LLM: {e}", exc_info=True)
    exit()

# --- Direct Call Test ---
try:
    logging.info("--- Starting Direct LLM Call Test (LangChain Wrapper) ---")
    
    # Correct message format for ChatModel.generate()
    test_messages: List[List[BaseMessage]] = [
        [HumanMessage(content="This is a test prompt to check for usage_metadata.")]
    ]
    
    # The .generate() method should return an LLMResult object
    result: LLMResult = llm.generate(test_messages)
    
    logging.info("Direct call completed.")
    
    # --- Check the llm_output field ---
    if result.llm_output:
        logging.info(f"Actual llm_output from direct call: {result.llm_output}")
    else:
        logging.info("Actual llm_output from direct call is None or empty.")

except Exception as e:
    logging.error(f"An error occurred during the direct call test: {e}", exc_info=True)
```

### Actual Behavior
When the script above is executed, the log output shows that llm_output is empty or None.
```md
2025-05-30 10:50:00,123 - INFO - Actual llm_output from direct call is None or empty.
```

(This matches the behavior observed in our more complex application using callbacks, where token counts are always zero because `usage_metadata` is never found).

### Expected Behavior
The `llm_output` field in the returned `LLMResult` object should be a dictionary containing the `usage_metadata` key, which holds the token counts from the Google API response.

For example, the expected output should look something like this:
```md
2025-05-30 10:50:00,123 - INFO - Actual llm_output from direct call: {'usage_metadata': {'prompt_token_count': 12, 'candidates_token_count': 5, 'total_token_count': 17}, 'prompt_feedback': ...}
```

### Additional Context
This issue seems to be specific to the `langchain-google-genai` integration layer.

I can confirm that when using the underlying `google-generativeai` SDK directly (bypassing LangChain), the `GenerateContentResponse` object correctly includes the ``usage_metadata``.

Here is a simple script demonstrating the direct SDK working as expected:
```python
import os
import google.generativeai as genai

load_dotenv()
genai.configure(api_key=os.getenv("GOOGLE_API_KEY"))

model = genai.GenerativeModel('gemini-1.5-flash')
response = model.generate_content("This is a direct SDK test.")

# This prints the usage metadata successfully
print(response.usage_metadata)
```

This proves that the data is available from the API but is being lost or omitted by the ``ChatGoogleGenerativeAI`` wrapper before it's passed into the ``LLMResult`` object. Thank you for looking into this!

hassaninnovate:
## Why

Token counts are attached to `AIMessage.usage_metadata` but never reach `LLMResult.llm_output`. Callback-based token tracking reads `llm_output` in `BaseCallbackHandler.on_llm_end`, so any such handler observes zero tokens for `ChatGoogleGenerativeAI`.

Reproduced on 4.3.4 with `gemini-2.5-flash`:

```
on_llm_end -> llm_output:             {}
on_llm_end -> message.usage_metadata: {'input_tokens': 8, 'output_tokens': 592, 'total_tokens': 600}
```

## What changed

Two sites are involved, and fixing only the first is not sufficient:

1. `_response_to_result` mirrors the counts it already computes for `AIMessage.usage_metadata` into `llm_output`, using the `token_usage` key convention shared with other LangChain chat integrations, alongside `usage_metadata` and `model_name`.

2. `BaseChatModel._combine_llm_outputs` returns `{}` by default, and `generate()` builds the returned `LLMResult` through it. Without an override the counts are still discarded for callers reading `LLMResult.llm_output` directly, so `ChatGoogleGenerativeAI` now overrides it to sum usage across generations.

After:

```
on_llm_end -> llm_output: {'token_usage': {'prompt_tokens': 8, 'completion_tokens': 642, 'total_tokens': 650}, ...}
generate()  -> llm_output: {'model_name': 'gemini-2.5-flash', 'token_usage': {...}}
```

## Areas needing careful review

- **Streaming semantics.** `_response_to_result` is shared with the streaming path, where `lc_usage` is a per-chunk delta via `subtract_usage`. The new `llm_output` therefore carries the same delta as `message.usage_metadata` on each chunk, keeping the two consistent. Streaming aggregation itself is deliberately unchanged.
- **Summing in `_combine_llm_outputs`** assumes one entry per prompt in a batch, matching `langchain-openai`.
- `llm_output` needed an explicit `dict[str, Any]` annotation; it was previously inferred as `dict[str, dict[str, Any]]` from its initializer.

## Tests

Four unit tests added in `tests/unit_tests/test_chat_models.py`, covering both sites plus the no-usage-metadata path. Full `tests/unit_tests` suite (377) passes; `ruff check`, `ruff format --check`, and `mypy` are clean.

Fixes #957.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -1295,7 +1295,7 @@ def _response_to_result(
     prev_usage: UsageMetadata | None = None,
 ) -> ChatResult:
     """Converts a Google AI response into a LangChain `ChatResult`."""
-    llm_output = (
+    llm_output: dict[str, Any] = (
         {"prompt_feedback": response.prompt_feedback.model_dump()}
         if response.prompt_feedback
         else {}
@@ -1347,6 +1347,19 @@ def _response_to_result(
             lc_usage = None
     except AttributeError:
         lc_usage = None
+
+    # `llm_output` is what `BaseCallbackHandler.on_llm_end` reads for token
+    # accounting. Without these keys, callback-based token trackers report zero
+    # even though the counts are available on `AIMessage.usage_metadata`.
+    if lc_usage is not None:
+        llm_output["token_usage"] = {
+            "prompt_tokens": lc_usage["input_tokens"],
+            "completion_tokens": lc_usage["output_tokens"],
+            "total_tokens": lc_usage["total_tokens"],
+        }
+        llm_output["usage_metadata"] = lc_usage
+    if response.model_version:
+        llm_output["model_name"] = response.model_version
 
     generations: list[ChatGeneration] = []
 
@@ -3353,6 +3366,36 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
             **request_params,
             **kwargs,
         )
+
+    def _combine_llm_outputs(
+        self, llm_outputs: list[dict[str, Any] | None], /
+    ) -> dict[str, Any]:
+        """Merge per-generation `llm_output` payloads for the final `LLMResult`.
+
+        `BaseChatModel` returns an empty dict by default, which would discard the
+        token counts populated in `_response_to_result` whenever a caller reads
+        `LLMResult.llm_output`. Summing the counts keeps a batched result
+        consistent with the per-generation values.
+
+        Args:
+            llm_outputs: Per-generation `llm_output` dicts, which may be `None`.
+
+        Returns:
+            The summed `token_usage` and the model name, when available.
+        """
+        combined: dict[str, Any] = {}
+        token_usage: dict[str, int] = {}
+        for llm_output in llm_outputs:
+            if not llm_output:
+                continue
+            for key, value in (llm_output.get("token_usage") or {}).items():
+                if value is not None:
+                    token_usage[key] = token_usage.get(key, 0) + value
+            if "model_name" in llm_output:
+                combined.setdefault("model_name", llm_output["model_name"])
+        if token_usage:
+            combined["token_usage"] = token_usage
+        return combined
 
     def _generate(
         self,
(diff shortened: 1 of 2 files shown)


hassaninnovate:
Closing this — it was opened from the wrong account. Reopening the identical change from the account that authored the commit.

PR #1962, opened 2026-08-25
fix(vertexai): bump pyarrow dependency

### LINKED ISSUE #1951: langchain-google-vertexai's pyarrow<24.0.0 conflicts with projects needing pyarrow 25
### Package (Required)

- [ ] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
[project]
name = "repro"
version = "0.1.0"
requires-python = ">=3.13"
dependencies = [
    "langchain-google-vertexai>=3.2.4",
    "pyarrow>=25.0.1",
]

Then `uv lock`.
```

```shell
× No solution found when resolving dependencies:
  ╰─▶ Because langchain-google-vertexai>=3.2.4 depends on
      pyarrow>=19.0.1,<24.0.0 and your project depends on
      langchain-google-vertexai>=3.2.4, we can conclude that your project
      depends on pyarrow>=19.0.1,<24.0.0.
      And because your project depends on pyarrow>=25.0.1, we can conclude
      that your project's requirements are unsatisfiable.
```

### Description

We ran into this in a monorepo where several services share an internal library. One service uses `langchain-google-vertexai`; another reads Parquet and needs pyarrow 25. Because the shared library declares `pyarrow`, vertexai's `<24.0.0` cap holds it back and the two can't coexist.

As far as I can tell vertexai is the only package in the tree that requires pyarrow, and doesn't import it directly – I assume it's there for the optional pyarrow paths in `google-cloud-bigquery` / `google-cloud-aiplatform`, which don't set an upper bound themselves.

Would you consider bumping the ceiling to `<26.0.0`?

**Environment:** Python 3.13 · langchain-google-vertexai 3.2.4 · uv 0.12.5

SitanshuA091:
## Why
The previous constraint:

```toml
pyarrow>=19.0.1,<24.0.0
```
prevented projects requiring `pyarrow>=25.0.0` from resolving dependencies alongside `langchain-google-vertexai`.

This change updates the dependency range to allow newer compatible `pyarrow` releases and regenerates the lock file with the updated dependency resolution.

## Changes
Updates the `pyarrow` dependency constraint in `libs/vertexai` to allow newer versions.
The existing upper bound was introduced as part of a previous dependency update. However, since `pyarrow` is not directly imported by `langchain-google-vertexai`, the restrictive upper bound caused unnecessary conflicts for downstream projects requiring newer `pyarrow` versions.

## Relevant issues

Fixes #1951

## Type

🚄 Infrastructure

## Changes(optional)

* Updated the `pyarrow` version constraint in `libs/vertexai/pyproject.toml`
* Regenerated `uv.lock` with the updated dependency resolution

## Testing(optional)

Verified locally:

* Installed package test dependencies
* Ran unit tests successfully:

```bash
uv run pytest tests/unit_tests -q
```
results - 300 passed

* Verified dependency lock update using:

```bash
uv lock --upgrade-package pyarrow
```

## Note(optional)

This PR only updates dependency constraints and the lock file. No runtime code changes were required.


mdrxy (approved):
Thanks for the fix. Confirmed pyarrow is not directly imported anywhere in `langchain_google_vertexai` — it's a transitive-only pin, so the restrictive `<24.0.0` upper bound was causing unnecessary resolver conflicts (reproduced in #1951). Widening to `<26.0.0` is safe; unit and integration tests pass. Approving.

(Reviewed with assistance from an AI agent.)

PR #1979, opened 2026-08-30
fix(genai): select thinking param by model family when both are set

### LINKED ISSUE #1462: `thinking_budget` is ignored when both `thinking_level` and `thinking_budget` are set for Gemini 2.x models
### Package (Required)

- [x] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python

# Use case: building a chat model that can work with both Gemini 2.x and 3.x
# by providing both thinking_budget (for 2.x) and thinking_level (for 3.x)
llm = ChatGoogleGenerativeAI(
    model="gemini-2.5-flash-preview-05-20",
    thinking_budget=1024,  # Should be used for Gemini 2.x models
    thinking_level="low",  # Should be used for Gemini 3+ models
)

# When using Gemini 2.x, thinking_budget should be applied
# but it is incorrectly ignored - thinking_level takes precedence regardless of model version
```

```shell
```

## Bug Description

In `_build_thinking_config()` method in `ChatGoogleGenerativeAI`, when both `thinking_level` and `thinking_budget` are provided, `thinking_budget` is always ignored with a warning that states `thinking_level` takes precedence "for Gemini 3+ models".

However, the actual code ignores `thinking_budget` **regardless of the model version**.

## Use Case

A valid use case is building a single chat model configuration that can work with both Gemini 2.x and Gemini 3.x models by providing both parameters:
- `thinking_budget` for Gemini 2.x models
- `thinking_level` for Gemini 3+ models

The code should automatically select the appropriate parameter based on the model version.

## The Problem

The warning message correctly states:
> "'thinking_level' takes precedence for Gemini 3+ models; 'thinking_budget' will be ignored."

But the code unconditionally prioritizes `thinking_level` over `thinking_budget` without checking the model version:

```python
# thinking_level takes precedence over thinking_budget for Gemini 3+ models
if thinking_level is not None:
    if thinking_budget is not None:
        warnings.warn(...)
    config["thinking_level"] = thinking_level
elif thinking_budget is not None:
    config["thinking_budget"] = thinking_budget
```

For Gemini 2.x models (where `thinking_budget` should be used instead of `thinking_level`), the `thinking_budget` parameter is incorrectly ignored when both are provided.

## Expected Behavior

The logic should check the model version and:
- For Gemini 3+ models: use `thinking_level`, warn if `thinking_budget` is also provided
- For Gemini 2.x models: use `thinking_budget`, warn if `thinking_level` is also provided (or ignore `thinking_level` silently)

## Code Location

`libs/genai/langchain_google_genai/chat_models.py` - `_build_thinking_config()` method (around line 2597-2630)

zamal-db:
**Description:** When both `thinking_level` and `thinking_budget` are set, `_build_thinking_config` always dropped `thinking_budget`, regardless of the model version. The accompanying warning claims `thinking_level` takes precedence "for Gemini 3+ models", but the code applied that precedence to every model, including Gemini 2.x where `thinking_level` is not supported.

This breaks the documented pattern of building a single model configuration that carries both parameters so it can target either family: `thinking_budget` for Gemini 2.x and `thinking_level` for Gemini 3+. On a Gemini 2.x model, `thinking_budget` was silently discarded and the request went out with an unsupported `thinking_level` instead.

**Reproduction (no API key needed):**

```python
import warnings
from pydantic import SecretStr

for model in ("gemini-2.5-flash", "gemini-3.5-flash"):
    llm = ChatGoogleGenerativeAI(
        model=model,
        google_api_key=SecretStr("k"),
        thinking_level="low",
        thinking_budget=128,
    )
    with warnings.catch_warnings(record=True):
        warnings.simplefilter("always")
        cfg = llm._prepare_request([HumanMessage(content="hi")])["config"].thinking_config
    print(f"{model:20s} level={cfg.thinking_level} budget={cfg.thinking_budget}")
```

Before this PR:

```
gemini-2.5-flash     level=ThinkingLevel.LOW budget=None   # thinking_budget wrongly dropped
gemini-3.5-flash     level=ThinkingLevel.LOW budget=None
```

After this PR:

```
gemini-2.5-flash     level=None budget=128                 # thinking_budget kept for 2.x
gemini-3.5-flash     level=ThinkingLevel.LOW budget=None   # thinking_level kept for 3+
```

**Fix:** When both parameters survive merging, keep the one matching the target model family using the existing `_is_gemini_3_or_later(self.model)` helper: `thinking_level` for Gemini 3+, `thinking_budget` for Gemini 2.x. The warning now names the parameter that was dropped and the family it applies to. Docstrings for `reasoning_effort`/`thinking_level` and `thinking_config` are updated to describe the version-aware selection.

**Issue:** Fixes #1462

**Dependencies:** None

**Changes:**

- `libs/genai/langchain_google_genai/chat_models.py`: version-aware selection in `_build_thinking_config`, plus docstring updates.
- `libs/genai/tests/unit_tests/test_chat_models.py`: added `test_thinking_budget_takes_precedence_for_gemini_2x`.

Existing `thinking_level` precedence tests (which use a Gemini 3+ model) still pass unchanged. Full `tests/unit_tests/test_chat_models.py` run: 409 passed.

### DIFF
--- a/libs/genai/langchain_google_genai/chat_models.py
+++ b/libs/genai/langchain_google_genai/chat_models.py
@@ -3058,7 +3058,8 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
     !!! note "Replaces `thinking_budget`"
 
         `thinking_budget` is deprecated for Gemini 3+ models. If both parameters are
-        provided, this field takes precedence.
+        provided, the one matching the model family is used: `thinking_level` for
+        Gemini 3+ models and `thinking_budget` for Gemini 2.x models.
 
         If left unspecified, the model's default thinking level is used.
 
@@ -3082,8 +3083,10 @@ class ChatGoogleGenerativeAI(_BaseGoogleGenerativeAI, BaseChatModel):
     !!! note "Precedence"
 
         If `thinking_config` is provided together with flat thinking arguments,
-        the flat arguments take precedence for matching fields. After merging,
-        `thinking_level` takes precedence over `thinking_budget` for Gemini 3+ models.
+        the flat arguments take precedence for matching fields. After merging, if
+        both `thinking_level` and `thinking_budget` are present, the one matching
+        the model family is used: `thinking_level` for Gemini 3+ models and
+        `thinking_budget` for Gemini 2.x models.
     """
 
     cached_content: str | None = None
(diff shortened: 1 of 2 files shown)


zamal-db:
Closing in favor of #1464, which already implements the same fix in _build_thinking_config (thinking_budget for Gemini 2.x, thinking_level for 3+) and predates this PR by several months. Deferring to the earlier PR.

Issue #1011, opened 2025-06-29
`UsageMetadata` for `ChatAnthropicVertex` is missing cache read/write token count in `input_token_details`


imranarshad:
### **Environment**
- **Package**: `langchain-google-vertexai`
- **Version**: 2.0.26 (latest official release)
- **Python Version**: 3.11+
- **Model**: `ChatAnthropicVertex` (Claude via Google Vertex AI)
- **Tested Model**: `claude-sonnet-4@20250514`

### **Problem Description**
`ChatAnthropicVertex` does not follow LangChain's `UsageMetadata` standard for cache token reporting. Instead of using the expected `input_token_details` structure, it either puts cache tokens at the top level (non-streaming) or omits them entirely (streaming). This breaks compatibility with LangChain's token usage patterns and the official `langchain_anthropic` package.

### **Expected Behavior**
`ChatAnthropicVertex` should follow LangChain's `UsageMetadata` standard by placing cache token information in `input_token_details`, consistent with:
1. **LangChain's UsageMetadata TypedDict definition**
2. **Official `langchain_anthropic.ChatAnthropic` implementation**
3. **Standard token usage patterns across LangChain**

### **Actual Behavior**
**⚠️ Verified with langchain-google-vertexai v2.0.26:**

**Streaming response** (missing cache tokens entirely):
```python
# First chunk from streaming response
{
    'input_tokens': 14,
    'output_tokens': 0,
    'total_tokens': 14
}
# ❌ Missing input_token_details completely
```

**Non-streaming response** (wrong cache token structure):
```python
{
    'input_tokens': 14,
    'output_tokens': 10,
    'total_tokens': 24,
    'cache_creation_input_tokens': 0,    # ❌ Wrong structure
    'cache_read_input_tokens': 0         # ❌ Should be in input_token_details
}
# ❌ Missing proper input_token_details structure
```

### **Expected Behavior**
**✅ With our fix implemented:**

**Streaming response** (cache tokens properly included):
```python
# First chunk from streaming response
{
    'input_tokens': 14,
    'output_tokens': 0,
    'total_tokens': 14,
    'input_token_details': {'cache_creation': 0, 'cache_read': 0}  # ✅ Correct
}
```

**Non-streaming response** (proper LangChain structure):
```python
{
    'input_tokens': 14,
    'output_tokens': 10,
    'total_tokens': 24,
    'input_token_details': {'cache_read': 0, 'cache_creation': 0}  # ✅ Correct
}
```

[...]

PR #1010, opened 2025-06-29
fix: Add cache token support to ChatAnthropicVertex streaming responses

imranarshad:
- Add reusable _create_usage_metadata() function matching langchain_anthropic
- Fix streaming message_start events to extract cache tokens
- Fix streaming message_delta events to follow official pattern
- Unify non-streaming to use same logic for consistency
- Use proper LangChain UsageMetadata structure with input_token_details
- Remove custom CacheUsageMetadata class in favor of standard types

Fixes missing cache token information in streaming responses to match official langchain_anthropic behavior and enable proper cost tracking.

## Type

🆕 New Feature
🐛 Bug Fix
🧹 Refactoring
📖 Documentation
🚄 Infrastructure
✅ Test

PR #1994, opened 2026-09-03
test(vertexai): cover cache tokens in Anthropic usage metadata

IshaanPotle:
Relates to #1011.

While looking at #1011 I found the reported behaviour is already correct on `main` — `_create_usage_metadata()` puts Anthropic cache counts in `input_token_details` and folds them into `input_tokens`. Running the issue's own repro gives exactly the output it asks for:

```python
{'input_tokens': 14, 'output_tokens': 10, 'total_tokens': 24,
 'input_token_details': {'cache_read': 0, 'cache_creation': 0}}
```

But nothing in `tests/unit_tests/` exercises that helper, so the fix isn't pinned. This PR adds three cases:

- cache fields present but zero → `input_token_details` still reported (the issue's repro; easy to regress by treating `0` as absent)
- non-zero cache counts → folded into `input_tokens` / `total_tokens`
- provider omits the cache fields → `input_token_details` omitted entirely

Tests only, no source changes. `#1011` can probably be closed once someone confirms the streaming side too — `message_start` routes through the same helper, and `message_delta` intentionally reports output tokens only, matching `langchain_anthropic`.

```
$ uv run pytest tests/unit_tests/test_anthropic_utils.py
65 passed in 1.44s

$ uv run ruff check tests/unit_tests/test_anthropic_utils.py
All checks passed!

$ uv run ruff format --check tests/unit_tests/test_anthropic_utils.py
1 file already formatted
```

---

AI disclaimer (per `AGENTS.md`): I used an AI coding assistant to help investigate the issue and draft these tests. I reviewed every line, confirmed the current behaviour by running the helper directly, and ran the checks above.

### DIFF
--- a/libs/vertexai/tests/unit_tests/test_anthropic_utils.py
+++ b/libs/vertexai/tests/unit_tests/test_anthropic_utils.py
@@ -8,6 +8,7 @@ from anthropic.types import (
     RawContentBlockDeltaEvent,
     SignatureDelta,
     ThinkingDelta,
+    Usage,
 )
 from langchain_core.messages import (
     AIMessage,
@@ -20,6 +21,7 @@ from langchain_core.messages.content import create_image_block, create_text_bloc
 from langchain_core.messages.tool import tool_call as create_tool_call
 
 from langchain_google_vertexai._anthropic_utils import (
+    _create_usage_metadata,
     _documents_in_params,
     _format_image,
     _format_message_anthropic,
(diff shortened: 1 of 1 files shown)


IshaanPotle:
Correcting my own framing here, since #1011 has now been closed as fixed by #1010.

I said "nothing in `tests/unit_tests/` exercises that helper". That's literally true — nothing calls `_create_usage_metadata` directly — but it was misleading, because #1010 already covers the behaviour indirectly through `test_anthropic_format_output` and `test_anthropic_format_output_with_chain_of_thoughts` in `test_chat_models.py`. I should have checked that before writing it.

So the honest scope of this PR is narrower than I implied. Of the three cases:

- **non-zero cache counts folded into `input_tokens`** — already covered by #1010. Redundant.
- **cache fields present but zero** — not covered anywhere; every existing test uses `1`. This is the issue's original repro, and it's the easy regression: a truthiness check like `if cache_read:` instead of `is not None` would silently drop `input_token_details` when a request simply had no cache activity.
- **cache fields absent entirely** — not covered; asserts `input_token_details` is omitted rather than emitted with `None`s.

Happy to go whichever way you prefer: keep it as-is, drop the redundant case and keep just the two boundary cases, or close this if you'd rather not add direct unit tests for a private helper. No attachment to it either way.

PR #2009, opened 2026-09-15
fix(vertexai): keep thinking blocks replayable when streaming from ChatAnthropicVertex

### LINKED ISSUE #2005: ChatAnthropicVertex: streaming drops content_block_start for thinking/redacted_thinking, producing thinking blocks with no thinking field that break subsequent turns
### Package (Required)

- [ ] langchain-google-genai

### Checked other resources

### Example Code (Python)

```python
from anthropic.types import (
    RawContentBlockStartEvent,
    RawContentBlockDeltaEvent,
    RawContentBlockStopEvent,
    SignatureDelta,
    TextBlock,
    TextDelta,
    ThinkingBlock,
)
from langchain_core.messages import AIMessage
from langchain_google_vertexai._anthropic_utils import (
    _make_message_chunk_from_anthropic_event,
    _format_message_anthropic,
)

# A thinking block that emits NO thinking_delta -- only a signature_delta.
# Happens with thinking={"type": "adaptive", "display": "summarized"} when the
# summary for that block is empty.
events = [
    RawContentBlockStartEvent(
        content_block=ThinkingBlock(thinking="", signature="", type="thinking"),
        index=0, type="content_block_start",
    ),
    RawContentBlockDeltaEvent(
        delta=SignatureDelta(signature="ErUBCkYIBxgCIkDx...", type="signature_delta"),
        index=0, type="content_block_delta",
    ),
    RawContentBlockStopEvent(index=0, type="content_block_stop"),
    RawContentBlockStartEvent(
        content_block=TextBlock(text="", type="text", citations=None),
        index=1, type="content_block_start",
    ),
    RawContentBlockDeltaEvent(
        delta=TextDelta(text="Here is the answer.", type="text_delta"),
        index=1, type="content_block_delta",
    ),
    RawContentBlockStopEvent(index=1, type="content_block_stop"),
]

aggregate = None
for event in events:
    chunk = _make_message_chunk_from_anthropic_event(
        event, stream_usage=True, coerce_content_to_string=False
    )
    if chunk is None:
        continue
    aggregate = chunk if aggregate is None else aggregate + chunk

print(aggregate.content)
print(_format_message_anthropic(AIMessage(content=aggregate.content), project=None))
print("'thinking' present?", "thinking" in aggregate.content[0])
```

```shell
[{'signature': 'ErUBCkYIBxgCIkDx...', 'type': 'thinking', 'index': 0},
 {'text': 'Here is the answer.', 'type': 'text', 'index': 1}]

{'role': 'assistant', 'content': [{'signature': 'ErUBCkYIBxgCIkDx...', 'type': 'thinking'},
                                  {'type': 'text', 'text': 'Here is the answer.'}]}

'thinking' present? False
```

### Package versions

| Package | Version |
| --- | --- |
| `langchain-google-vertexai` | `3.2.4` |
| `langchain-core` | `1.6.2` |
| `anthropic` | `1.5.0` |
| Python | `3.11.15` (macOS arm64) |

### Description

When streaming from `ChatAnthropicVertex` with extended thinking enabled, `_make_message_chunk_from_anthropic_event` only emits a chunk for `content_block_start` when the block is a `tool_use`. Thinking blocks are therefore reconstructed **entirely from delta events**.

If a thinking block emits a `signature_delta` but **no `thinking_delta`**, the aggregated content block ends up as:

```python
{"type": "thinking", "signature": "...", "index": 0}
```

— a thinking block with **no `thinking` key at all**.

`_format_message_anthropic` then passes that straight back to the API on the next turn, and the request is rejected:

```
BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error',
'message': 'messages.93.content.0.thinking.thinking: Field required'}}
```

This is reliably triggered by `thinking={"type": "adaptive", "display": "summarized"}`, where Anthropic always produces a signature for every thinking block but only emits `thinking_delta`s when there is summary text to show. On low-effort hops (tool-call steps, short turns) the summary is empty, so zero `thinking_delta`s arrive.

The failure is **sticky**: once such a message is stored in conversation history, every subsequent turn replays the malformed block and fails, so the conversation becomes permanently unusable. This is why it only surfaces in long multi-turn conversations, and why the reported index is deep in the message list.

### Reproduction

No credentials or network needed — this drives the event handler directly with real `anthropic` SDK types.

```python
from anthropic.types import (
    RawContentBlockStartEvent,
    RawContentBlockDeltaEvent,
    RawContentBlockStopEvent,
    SignatureDelta,
    TextBlock,
    TextDelta,
    ThinkingBlock,
)
from langchain_core.messages import AIMessage
from langchain_google_vertexai._anthropic_utils import (
    _make_message_chunk_from_anthropic_event,
    _format_message_anthropic,
)

# A thinking block that emits NO thinking_delta -- only a signature_delta.
# Happens with thinking={"type": "adaptive", "display": "summarized"} when the
# summary for that block is empty.
events = [
    RawContentBlockStartEvent(
        content_block=ThinkingBlock(thinking="", signature="", type="thinking"),
        index=0,
        type="content_block_start",
    ),
    RawContentBlockDeltaEvent(
        delta=SignatureDelta(signature="ErUBCkYIBxgCIkDx...", type="signature_delta"),
        index=0,
        type="content_block_delta",
    ),
    RawContentBlockStopEvent(index=0, type="content_block_stop"),
    RawContentBlockStartEvent(
        content_block=TextBlock(text="", type="text", citations=None),
        index=1,
        type="content_block_start",
    ),
    RawContentBlockDeltaEvent(
        delta=TextDelta(text="Here is the answer.", type="text_delta"),
        index=1,
        type="content_block_delta",
    ),
    RawContentBlockStopEvent(index=1, type="content_block_stop"),
]

aggregate = None
for event in events:
    chunk = _make_message_chunk_from_anthropic_event(
        event, stream_usage=True, coerce_content_to_string=False
    )
    if chunk is None:
        continue
    aggregate = chunk if aggregate is None else aggregate + chunk

print("aggregated AIMessageChunk.content:")
print(" ", aggregate.content)

payload = _format_message_anthropic(AIMessage(content=aggregate.content), project=None)
print("\npayload sent back to the API on the next turn:")
print(" ", payload)

print("\n'thinking' key present?", "thinking" in payload["content"][0])
```

### Actual output

```
aggregated AIMessageChunk.content:
  [{'signature': 'ErUBCkYIBxgCIkDx...', 'type': 'thinking', 'index': 0},
   {'text': 'Here is the answer.', 'type': 'text', 'index': 1}]

payload sent back to the API on the next turn:
  {'role': 'assistant', 'content': [{'signature': 'ErUBCkYIBxgCIkDx...', 'type': 'thinking'},
                                    {'type': 'text', 'text': 'Here is the answer.'}]}

'thinking' key present? False
```

Sending that assistant message back produces `...content.0.thinking.thinking: Field required`.

### Expected output

The thinking block should retain its `thinking` field (an empty string is valid), so the message can be replayed:

```
[{'type': 'thinking', 'thinking': '', 'signature': 'ErUBCkYIBxgCIkDx...', 'index': 0}, ...]
```

### Root cause

`langchain_google_vertexai/_anthropic_utils.py`, `_make_message_chunk_from_anthropic_event`:

1. **Lines 564-568** — `content_block_start` is only handled for `tool_use`:

   ```python
   elif (
       event.type == "content_block_start"
       and event.content_block is not None
       and event.content_block.type == "tool_use"
   ):
   ```

   So the opening `ThinkingBlock` (which always carries a `thinking` field) is discarded.

2. **Lines 593-599** — thinking blocks are rebuilt from deltas alone:

   ```python
   elif event.delta.type in {"thinking_delta", "signature_delta"}:
       content_block = event.delta.model_dump()
       if "text" in content_block and content_block["text"] is None:
           content_block.pop("text")
       content_block["index"] = event.index
       content_block["type"] = "thinking"
       message_chunk = AIMessageChunk(content=[content_block])
   ```

   `SignatureDelta.model_dump()` yields only `{"type", "signature"}`, so a signature-only block never gains a `thinking` key.

3. **Lines 203-210** — `_format_message_anthropic` key-filters thinking blocks to `("type", "thinking", "cache_control", "signature")` but injects no default, so the malformed block is serialised as-is:

   ```python
   if block["type"] == "thinking":
       content.append(
           {
               k: v
               for k, v in block.items()
               if k in ("type", "thinking", "cache_control", "signature")
           }
       )
       continue
   ```

### Related: `redacted_thinking` is dropped entirely

`redacted_thinking` blocks are delivered **only** via `content_block_start` (they have no deltas). Because that branch is `tool_use`-only, they are silently discarded from the aggregated message. With thinking enabled this both loses data and can cause downstream validation errors on replay.

### This was already fixed in `langchain-anthropic`

`langchain-anthropic` 1.7.2 handles `content_block_start` for `text`, `thinking`, `redacted_thinking`, `tool_use`, `tool_result` and `document` (`langchain_anthropic/chat_models.py`, approx. L2010-2080), with explicit regression tests:

- `test_thinking_content_block_start_carries_initial_thinking`
- `test_empty_thinking_content_block_start_emits_no_chunk`
- `test_text_content_block_start_carries_initial_text`

Its docstring describes exactly this class of bug:

> The handler previously only built a chunk on `content_block_start` for tool / document / redacted_thinking blocks, so the leading thinking content was silently dropped from the aggregated message.

The relevant upstream branch:

```python
elif (
    event.type == "content_block_start"
    and event.content_block is not None
    and event.content_block.type in ("text", "thinking")
):
    # Anthropic can place the opening content of a text or thinking block
    # directly on the `content_block_start` event instead of in a
    # following delta. ...
    block_start_event = event
    if event.content_block.type == "text":
        ...
    else:  # thinking
        thinking = getattr(event.content_block, "thinking", "") or ""
        signature = getattr(event.content_block, "signature", "") or ""
        if thinking or signature:
            content_block = event.content_block.model_dump()
            content_block["index"] = event.index
            content_block["type"] = "thinking"
            message_chunk = AIMessageChunk(content=[content_block])
```

`ChatAnthropicVertex` maintains its own copy of this logic in `_anthropic_utils.py` and has not received the fix.

### Suggested fix

Port the `langchain-anthropic` behaviour into `_anthropic_utils.py`:

1. Extend the `content_block_start` branch to also handle `thinking`, `redacted_thinking` and `text`, emitting `event.content_block.model_dump()` (with `index`) when it carries content, so the block always starts life with a `thinking` key.
2. Defensively, in `_format_message_anthropic`, default `thinking` to `""` when serialising a `type: "thinking"` block that lacks it, so pre-existing persisted histories don't hard-fail.

Item 2 matters in practice: malformed blocks produced by earlier versions are already stored in users' conversation histories and will keep failing until they're normalised.

revanth-045:
Fixes #2005

## Problem

`_make_message_chunk_from_anthropic_event` only emits a chunk for `content_block_start` when the block is a `tool_use`. Two consequences:

1. **Thinking blocks are rebuilt purely from deltas.** A block that emits a `signature_delta` but no `thinking_delta` — which is exactly what summarized adaptive thinking (`thinking={"type": "adaptive", "display": "summarized"}`) produces whenever a block's summary is empty — aggregates to `{"type": "thinking", "signature": "...", "index": 0}` with **no `thinking` field**. `_format_message_anthropic` replays it as-is, and the API rejects the next turn with `messages.N.content.0.thinking.thinking: Field required`. Because the malformed block is now in history, every subsequent turn fails too.
2. **`redacted_thinking` blocks are dropped entirely** — they carry their whole payload on the start event and never emit deltas.

## Fix

Mirror `langchain-anthropic`:
- emit the `content_block_start` chunk for `thinking` (with `thinking` defaulted to `""`) and for `redacted_thinking`;
- `setdefault("thinking", "")` on `signature_delta`, so the required field exists even if the start event was somehow not seen.

`AIMessageChunk` addition merges the following deltas into the opened block by `index`, exactly as before, so blocks that *do* stream `thinking_delta`s are unchanged apart from now also preserving any thinking text placed on the start event.

## Tests

- `test_thinking_block_with_only_signature_keeps_thinking_field` — the issue's exact event sequence, driven through aggregation and `_format_message_anthropic`; asserts the replayed block carries `thinking: ""`. Fails on `main`.
- `test_thinking_block_start_content_is_not_dropped` — thinking on the start event is kept and merged with later deltas. Fails on `main`.
- `test_redacted_thinking_block_start_is_emitted`. Fails on `main`.
- Existing `test_make_thinking_message_chunk_from_anthropic_event` updated: a signature chunk now also carries `thinking: ""`.

`make lint` (ruff, isort, mypy) clean; `test_anthropic_utils.py` + `test_chat_models.py` 180 passed.


aii-vtl:
Thank you for the PR, I would suggest reviewing and fixing this ASAP as it's a MAJOR issue with the library, borderline making the whole library impossible to use.

Issue #2008, opened 2026-09-15
ChatAnthropicVertex: temperature/top_p/top_k raise TypeError with anthropic SDK v1


KyriosGN0:
### Summary

`ChatAnthropicVertex` cannot be used with the `anthropic` SDK v1. Any call that has `temperature`, `top_p`, or `top_k` set raises `TypeError` before the request leaves the process, for both `invoke` and `stream`.

The v1 SDK removed those three parameters from `messages.create()` ([migration guide](https://github.com/anthropics/anthropic-sdk-python/blob/main/MIGRATION.md#removed-deprecated-request-parameters)); they must now go through `extra_body`. `ChatAnthropicVertex` still passes them as named arguments.

### Repro

```python
from langchain_google_vertexai.model_garden import ChatAnthropicVertex

m = ChatAnthropicVertex(
    model_name="claude-sonnet-4-5@20250929",
    project="<your-project>",
    location="us-east5",
    temperature=0.1,
)
m.invoke("say hi")
```

```
TypeError: Messages.create() got an unexpected keyword argument 'temperature'
```

`list(m.stream("say hi"))` fails the same way. Without `temperature` (and `top_p`/`top_k`) the call succeeds, because `_format_params` drops `None` values — so the failure only shows up once a sampling parameter is actually set.

### Cause

`_default_params` (`libs/vertexai/langchain_google_vertexai/model_garden.py`) always includes `temperature`, `top_k` and `top_p`, and the resulting dict is passed straight to `self.client.messages.create(**params)` / `self.async_client.messages.create(**params)`.

### Environment

- `langchain-google-vertexai` 3.2.4
- `anthropic` 1.5.0
- Python 3.11

### Notes

- The `anthropic` extra pins `anthropic>=0.35.0,<1.0.0`, which does not protect projects that depend on `anthropic` directly rather than on `langchain-google-vertexai[anthropic]` — there the resolver is free to pick v1 and the break only appears at runtime. `langchain-anthropic` already allows `anthropic<2.0.0`.
- One more v1 change that touches this file: the SDK's HTTP layer moved from `httpx` to `httpx2`, so the `http_client` / `async_http_client` objects forwarded to `AnthropicVertex` in `validate_environment` must be `httpx2` clients once v1 is supported. Passing an old `httpx.Client` raises `TypeError`.

### Suggested fix

[...]

PR #2012, opened 2026-09-16
fix(vertexai): route ChatAnthropicVertex sampling params through extra_body

AmirF194:
`_default_params` always includes `temperature`/`top_k`/`top_p` and hands them straight to `messages.create(**params)`. anthropic 1.x dropped those from `create()`'s own signature (confirmed by reading `anthropic==1.5.0`'s `Messages.create`), so any call with a sampling param set raises `TypeError: Messages.create() got an unexpected keyword argument 'temperature'` before the request leaves the process.

`extra_body` is merged into the request body on both 0.35-0.99 and 1.x (`_base_client._build_request` merges `options.extra_json` into `json_data`), so I moved the three sampling params there instead, the same way `_move_betas_to_extra_body` already does for `betas`. Added `_move_sampling_params_to_extra_body` and call it at the same four `.create()` sites.

Left the extra's `<1.0.0` upper bound alone: the issue also flags that `http_client`/`async_http_client` need to be `httpx2` clients once v1 is supported, and I didn't touch that, so bumping the pin here would still break for anyone passing a custom `http_client`. Filing this as `Refs #2008` rather than `Fixes` since that part is still open.

Verified: `pytest tests/unit_tests/` passes (302, was 300 on main), `ruff check`/`ruff format --diff`/`mypy` clean on `langchain_google_vertexai` and `tests`. Confirmed the `TypeError` against real `anthropic==1.5.0` (`Messages.create()` has no `temperature`/`top_p`/`top_k` param at all in that version) and that passing the same values via `extra_body` gets past kwarg binding instead. Anthropic Vertex itself I can't reach from here (no GCP project), so the fix is unit-tested on the dict transform, not run end to end against a live model.

AI disclaimer (per `AGENTS.md`): I used an AI coding assistant to investigate this and draft the fix. I reviewed every line, confirmed the signature change against the real package versions above, and ran the checks myself.


AmirF194:
Bumping this after a week of silence. The one red check is the Vertex integration suite, which needs live GCP credentials a fork PR never gets, so that's not something on my end to fix.

AmirF194:
#2016 fixes the same v1 SDK sampling-param issue here, and it is more careful than mine: it inspects the actual messages_api signature before moving params, so it stays correct if a future SDK version restores them, where mine moves unconditionally. This one was open first, but happy to close in favor of #2016 if you'd rather take that approach.

PR #2016, opened 2026-09-18
fix(vertexai): support Anthropic v1 sampling parameters

### LINKED ISSUE #2008: ChatAnthropicVertex: temperature/top_p/top_k raise TypeError with anthropic SDK v1

spark-328:
## Summary

- detect whether the active Anthropic messages API accepts `temperature`, `top_p`, and `top_k`
- when those named parameters are absent, forward their values through `extra_body` instead of passing unsupported keyword arguments
- apply the adapter consistently to sync, async, streaming, and beta request paths
- preserve the existing top-level behavior for legacy Anthropic clients

This makes the sampling parameters usable with Anthropic SDK v1-shaped clients while keeping the request body equivalent for older clients. It does not change the optional dependency bounds or the transport type expected by a caller-provided HTTP client.

## Verification

- `python -m uv run --group test pytest tests/unit_tests -q` (`303 passed, 1 skipped`)
- `ruff check` on the changed files passed
- `ruff format --check` on the changed files passed
- `mypy` on the changed files passed

Fixes #2008


spark-328:
The only red check is the Vertex `llm-integration-tests` Cloud Build. The sync, async, streaming, and beta regression tests pass locally (`303 passed, 1 skipped`), along with Ruff and mypy. Could someone with access rerun that external integration job when convenient, @ccurme?

